PR #23, opened 2017-04-12
generate no content for 0 amount transfers

JordanValdma:


src/main/java/ee/tuleva/onboarding/mandate/content/HtmlMandateContentCreator.java
```diff
@@ -78,7 +79,9 @@ private MandateContentFile getFutureContributionsFundMandateContentFile(Mandate
     private Map<String, List<FundTransferExchange>> getPrintableFundExchangeStructure(Mandate mandate) {
         Map<String, List<FundTransferExchange>> exchangeMap = new HashMap<>();
 
-        mandate.getFundTransferExchanges().stream().forEach(exchange -> {
+        mandate.getFundTransferExchanges().stream()
+                .filter(fte -> fte.getAmount().compareTo(BigDecimal.ZERO) == 1)
```
ErkoRisthein: Maybe extract it into a separate method to make it more readable. Something like:
```
private boolean biggerThan(BigDecimal value1, BigDecimal value2) {
    return value1.compareTo(value2) == 1;
}
```

PR #28, opened 2017-04-18
Member authentication refactoring

JordanValdma:


src/main/java/ee/tuleva/onboarding/auth/authority/GrantedAuthorityFactory.java
```diff
@@ -12,7 +12,7 @@
 
     public static Collection<GrantedAuthority> from(AuthenticatedPerson authenticatedPerson) {
 
-        if(authenticatedPerson.getUser().isPresent()) {
+        if(authenticatedPerson.getUser().getMember().isPresent()) {
```
ErkoRisthein: Could also use `.map().orElse()`, right?

PR #29, opened 2017-04-20
Creates a user in the database, when one is not present. Makes users.…

JordanValdma:
…email field nullable


src/main/java/ee/tuleva/onboarding/user/User.java
```diff
@@ -63,4 +62,14 @@ public int getAge() {
...
+	@PrePersist
+	protected void onCreate() {
+		createdDate = Instant.now();
+		updatedDate = Instant.now();
+	}
+
+	@PreUpdate
+	protected void onUpdate() {
+		updatedDate = Instant.now();
+	}
```
ErkoRisthein: Nice!

PR #43, opened 2017-05-04
Added minimum age requirement: 18

ErkoRisthein:


JordanValdma:
Should we filter under aged in login or payment or before the payment?

ErkoRisthein:
Ideally, as soon as possible. So that would be login. Do we validate the user on login somehow right now? If not, then they would hit the validation wall on the signup form.

PR #63, opened 2017-07-03
Updates user name on incoming payment request

ErkoRisthein:


src/main/java/ee/tuleva/onboarding/user/UserService.java
```diff
@@ -67,4 +71,15 @@ private User updateUser(User user) {
     mailChimpService.createOrUpdateMember(savedUser);
     return savedUser;
   }
+
+  public User updateNameIfMissing(User user, String fullName) {
```
JordanValdma: why not always update?

ErkoRisthein: Some risks that I see in doing that:
- maybe they paid from the wrong account (not in their name)
- backward compatible with the old flow, where we take the name from their ID-card / Mobile-ID (and which are definitely more accurate sources for their name)
- the full name splitting algorithm might yield incorrect results if you have more than 1 last name (not very likely for Estonians, but might be the case for foreigners)
- I think it's safer to not overwrite data if it already exists - i.e. if an existing member _somehow_ manages to get this method called on their user (via a bug of some sort) it wouldn't overwrite any existing data

PR #72, opened 2017-07-19
Adds member total count HEAD request. Adds additional check for clien…

JordanValdma:
…t credentials access to POST /users only


src/main/java/ee/tuleva/onboarding/member/MemberController.java
```diff
@@ -0,0 +1,38 @@
...
+@Slf4j
+@RequiredArgsConstructor
+@RestController
+@RequestMapping("/v1")
+public class MemberController {
+
+    private final MemberRepository memberRepository;
+
+    @ApiOperation(value = "Get total count of members")
+    @RequestMapping(method = HEAD, value = "/members")
```
ErkoRisthein: Nice and RESTful :beer:

PR #78, opened 2017-07-25
Adds audit log

JordanValdma:


src/main/java/ee/tuleva/onboarding/account/AccountStatementService.java
```diff
@@ -35,6 +38,7 @@
     @Cacheable(value=ACCOUNT_STATEMENT_CACHE_NAME, key="#person.personalCode")
     public List<FundBalance> getMyPensionAccountStatement(Person person) {
         log.info("Getting pension account statement for {} {}", person.getFirstName(), person.getLastName());
+        auditEventPublisher.publish(person.getPersonalCode(), AuditEventType.GET_ACCOUNT_STATEMENT);
```
ErkoRisthein: Would it make sense to use AOP instead to publish audit events?
For example: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-samples/spring-boot-sample-aop/src/main/java/sample/aop/monitor/ServiceMonitor.java

src/main/java/ee/tuleva/onboarding/audit/AuditServiceMonitor.java
```diff
@@ -0,0 +1,21 @@
...
+import ee.tuleva.onboarding.auth.principal.Person;
+import lombok.RequiredArgsConstructor;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Before;
+import org.springframework.stereotype.Component;
+
+@Aspect
+@Component
+@RequiredArgsConstructor
+public class AuditServiceMonitor {
```
ErkoRisthein: 👊

PR #87, opened 2018-07-17
Start adding fund comparison calculation using IRR

Tankenstein:
Also deletes the old unused comparison calculation.


src/test/groovy/ee/tuleva/onboarding/comparisons/FundComparisonCalculatorServiceSpec.groovy
```diff
@@ -0,0 +1,174 @@
...
+                accountOverviewProvider,
+                estonianAverageValueProvider,
+                marketAverageValueProvider
+        )
+    }
+
+    def "it uses an account overview for the given person and start time" () {
+        given:
+            Person person = Mock(Person)
+            Instant startTime = Instant.ofEpochSecond(1531813018) // 17 July 2018
```
oliverviljamaa: Let's use `parseInstant` here as well for readability. 🙂

PR #93, opened 2018-07-19
This finishes the comparison calculator.

Tankenstein:
Adds a controller, adds a fundprovider to get estonian funds in the calculation, adds integration to epis.


src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
```diff
@@ -0,0 +1,47 @@
...
+@RestController
+@RequestMapping("/v1")
+@RequiredArgsConstructor
+public class FundComparisonController {
+
+    private static final Instant DEFAULT_TIME = parseInstant("1900-01-01");
+
+    private final FundComparisonCalculatorService fundComparisonCalculatorService;
+
+    @ApiOperation(value = "Compare the current user's return to estonian and world average")
```
oliverviljamaa: That makes it seem like it expects a return as an input. How about `Gets current user's, Estonian average, and world average return percentages`?

src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
```diff
@@ -0,0 +1,47 @@
...
+    @GetMapping("/fund-comparison")
+    public FundComparison getComparison(
+            @ApiIgnore @AuthenticationPrincipal AuthenticatedPerson authenticatedPerson,
+            @RequestParam(value = "from", required = false) @DateTimeFormat(pattern="yyyy-MM-dd") Date startDate) {
```
oliverviljamaa: Is this a standard date time format for APIs?

src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
Tankenstein: Does not get world average

src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
Tankenstein: iso strings are but internally we have seemed to use this

PR #99, opened 2018-07-20
Run fund value retrieval on startup

Rauno56:


Rauno56:
Those changes got merged in another PR #100 .

PR #125, opened 2019-05-13
User residency field

v3rm0n:


src/main/java/ee/tuleva/onboarding/user/User.java
```diff
@@ -43,6 +43,8 @@
 
     private String lastName;
 
+    private Boolean resident;
```
ErkoRisthein: `isResident`?

src/main/resources/db/migration/V1_16__user_resident.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+  ADD COLUMN resident BOOLEAN default null;
```
ErkoRisthein: `is_resident`?

PR #126, opened 2019-05-13
Added attributes to fund:

mativait:
* pillar
 * ongoing_charges_figure


src/main/resources/db/migration/V1_18__add_pillar_attribute_to_fund.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE fund ADD COLUMN pillar SMALLINT;
```
ErkoRisthein: `NOT NULL`

src/main/resources/db/migration/V1_18__add_pillar_attribute_to_fund.sql
```diff
@@ -0,0 +1,2 @@
...
+UPDATE fund SET pillar = 2 where pillar is null;
```
ErkoRisthein: `ALTER TABLE fund ALTER COLUMN pillar SET NOT NULL;`

src/main/resources/db/migration/V1_17__add_ongoing_charge_figure_to_fund.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE fund
+  ADD COLUMN ongoing_charges_figure NUMERIC(10, 8) NOT NULL DEFAULT 0;
```
ErkoRisthein: Instead of having it default to 0, it's probably better to 
1. add column
2. set values to 0
3. add not null constraint

PR #136, opened 2019-05-15
Adding pillar for Account Statement

alarmatwork:


src/main/java/ee/tuleva/onboarding/account/AccountStatementController.java
```diff
@@ -22,14 +33,30 @@
...
+    private final EpisAccountOverviewProvider episAccountOverviewProvider;
+    private static final Instant START_TIME = Utils.parseInstant("2002-01-01");
 
     @ApiOperation(value = "Get pension register account statement")
     @RequestMapping(method = GET, value = "/pension-account-statement")
     public List<FundBalance> getMyPensionAccountStatement(@ApiIgnore @AuthenticationPrincipal AuthenticatedPerson authenticatedPerson,
                                                           @RequestHeader(value = "x-statistics-identifier", required = false) UUID statisticsIdentifier) {
         List<FundBalance> fundBalances = accountStatementService.getAccountStatement(authenticatedPerson);
+
+        //Calculate total contribution for each fund balance
```
ErkoRisthein: This should move to the `Service` layer - not to be done in the `Controller`

ErkoRisthein: i.e. in the `AccountStatementService`

ErkoRisthein: Also, no test for this `contributionSum` calculation logic :D

src/main/java/ee/tuleva/onboarding/common/Utils.java
```diff
@@ -0,0 +1,17 @@
...
+import java.text.SimpleDateFormat;
+import java.time.Instant;
+
+public class Utils {
+
+    public static Instant parseInstant(String format) {
+        try {
+            return new SimpleDateFormat("yyyy-MM-dd").parse(format).toInstant();
+        } catch (ParseException e) {
+            e.printStackTrace();
```
ErkoRisthein: Prints to standard out, better to avoid (not logged to file).
Not sure if it makes sense to catch the `ParseException` and return `Instant.now()` in that case?
I'd rather fail fast instead of silently.

ErkoRisthein: If we remove the `catch` block then maybe we won't need the `Utils` class at all - we can just use `new SimpleDateFormat("yyyy-MM-dd").parse(format).toInstant();` inline where needed.

src/main/java/ee/tuleva/onboarding/account/AccountStatementController.java
```diff
@@ -27,7 +38,7 @@
     @RequestMapping(method = GET, value = "/pension-account-statement")
     public List<FundBalance> getMyPensionAccountStatement(@ApiIgnore @AuthenticationPrincipal AuthenticatedPerson authenticatedPerson,
                                                           @RequestHeader(value = "x-statistics-identifier", required = false) UUID statisticsIdentifier) {
-        List<FundBalance> fundBalances = accountStatementService.getAccountStatement(authenticatedPerson);
+        List<FundBalance> fundBalances = accountStatementService.getAccountStatement(authenticatedPerson, true);
```
ErkoRisthein: I'd avoid boolean method arguments whenever possible.

PR #150, opened 2019-05-17
Get pillar from fund

v3rm0n:


src/main/java/ee/tuleva/onboarding/mandate/command/CreateMandateCommandToMandateConverter.java
```diff
@@ -52,11 +49,35 @@ public Mandate convert(CreateMandateCommandWithUser createMandateCommandWithUser
         return mandate;
     }
 
+    private Integer getPillar(CreateMandateCommand createMandateCommand) {
```
ErkoRisthein: Extract to `fundService`

PR #157, opened 2019-05-22
Fix capital rounding bug

ErkoRisthein:


src/main/java/ee/tuleva/onboarding/capital/CapitalService.java
```diff
@@ -5,64 +5,54 @@
...
+            getCapitalAmount(events, UNVESTED_WORK_COMPENSATION),
+            getCapitalAmount(events, WORK_COMPENSATION),
+            getProfit(events)
+        );
     }
 
     @NotNull
     private BigDecimal getCapitalAmount(List<MemberCapitalEvent> events, MemberCapitalEventType capitalPayment) {
         return events.stream().filter(event -> event.getType() == capitalPayment)
-            .filter(event -> event.getAccountingDate().compareTo(LocalDate.now()) < 1)
```
v3rm0n: I would create a method here
```
Boolean isBeforeOrEqual(LocalDate first, LocalDate second){
return first.isBefore(second) || first.isEqual(second);
}
```

ErkoRisthein: Equal is probably an edge case (i.e. it has to be equal by the nanosecond - highly unlikely). I'll just use `isBefore()` for better readability.

ErkoRisthein: Scratch that, LocalDate has no time :)

PR #173, opened 2019-08-29
Added the oauth2 rest template for epis calls with the client_credent…

ErkoRisthein:
…ials token

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/config/Oauth2RestTemplateConfiguration.java
+++ b/src/main/java/ee/tuleva/onboarding/config/Oauth2RestTemplateConfiguration.java
@@ -3,14 +3,11 @@ package ee.tuleva.onboarding.config;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.security.oauth2.client.OAuth2ClientContext;
 import org.springframework.security.oauth2.client.OAuth2RestOperations;
 import org.springframework.security.oauth2.client.OAuth2RestTemplate;
 import org.springframework.security.oauth2.client.token.grant.client.ClientCredentialsResourceDetails;
-import org.springframework.security.oauth2.config.annotation.web.configuration.EnableOAuth2Client;
 
 @Configuration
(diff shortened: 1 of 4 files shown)


ErkoRisthein:
Example usage:
```
    private final OAuth2RestOperations oAuth2RestTemplate;

    @Value("${epis.service.url}")
    private String episServiceUrl;

    ...

    ResponseEntity<FundDto[]> response = oAuth2RestTemplate.getForEntity(episServiceUrl + "/funds", FundDto[].class);
```

src/main/java/ee/tuleva/onboarding/config/Oauth2RestTemplateConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+public class Oauth2RestTemplateConfiguration {
+
+    @Bean
+    @ConfigurationProperties("epis.service.client")
+    ClientCredentialsResourceDetails oauth2ClientCredentialsResourceDetails() {
+        return new ClientCredentialsResourceDetails();
+    }
+
+    @Bean
+    public OAuth2RestOperations oauth2RestTemplate(ClientCredentialsResourceDetails resource, OAuth2ClientContext context) {
```
v3rm0n: I'm wondering if `EnableOAuth2Client` and `OAuth2ClientContext` are actually needed here 🤔 Doesn't it work without them?

ErkoRisthein: Not sure, but it's taken from the official spring javadocs: https://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/config/annotation/web/configuration/EnableOAuth2Client.html

v3rm0n: This is for Authorization Code Grant, which definitely needs oauth context.

ErkoRisthein: You're right, it does work without the `@EnableOAuth2Client` and `OAuth2ClientContext` here. 🙇

PR #213, opened 2020-03-12
Update docker script

ristosaar:
fix for postgres image won't start.
Seems to be a breaking change introduced here https://github.com/docker-library/postgres/pull/658

Should we add a database password in the future instead?


v3rm0n:
We should probably specify the version for the postgres container as well, since it is probably out of sync with our actual DB anyway, which is 9.6.12 at the moment....

PR #214, opened 2020-03-17
Id card authentication through OCSP

ristosaar:
Replacement module for deprecated DigiDocService.
-ID card certificate validity check through OCSP service
-ID card certificate expiry check


src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public String checkCertificate(X509Certificate certificate){
+        if(hasCertificateExpired(certificate)){
+            return "CERTIFICATE_EXPIRED";
+        }
+        return checkSerialNumber(certificate.getSerialNumber());
+    }
+
+    private boolean hasCertificateExpired(X509Certificate certificate) {
+        return System.currentTimeMillis() > certificate.getNotAfter().getTime();
```
v3rm0n: Using time this way is flaky, can fail if there are timezone differences. Try to use an instance of Clock instead.

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+    private final String method;
+
+    public OCSPRequest(String caToCheck, String ocspServer) {
+        this.ca = caToCheck;
+        this.ocspServer = ocspServer;
+        this.method = "POST";
+    }
+
+    public String checkCertificate(X509Certificate certificate){
+        if(hasCertificateExpired(certificate)){
```
ErkoRisthein: Unformatted code. Maybe we should introduce an automatic code style checker to the Gradle build? @v3rm0n can you recommend any good tools for Java & Gradle?

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    private boolean hasCertificateExpired(X509Certificate certificate) {
+        return System.currentTimeMillis() > certificate.getNotAfter().getTime();
+    }
+
+    public String checkSerialNumber(BigInteger serialNumber) {
+
+        String checkedResponse = null;
+        Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
```
v3rm0n: This should probably be done only once per app run

ErkoRisthein: Import `BouncyCastleProvider`

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        try {
+            log.info("Generating and sending OCSPRequest");
+            OCSPResp response = getOCSPResponse(serialNumber);
+
+            log.info("Validating OCSPResponse");
+            checkedResponse = validateOCSPResponse(response);
```
ErkoRisthein: I would return directly from here, instead of:
- defining a _mutable_ `String checkedResponse = null;` variable
- assigning to that variable
- returning that variable in the end

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+
+            request = ocspGen.build();
+        } catch (CertificateException | OCSPException | OperatorCreationException | UnsupportedEncodingException e) {
+            throw new AuthenticationException(AuthenticationException.Code.UNABLE_TO_TEST_USER_CERTIFICATE, "Uncaught error", e);
+        }
+
+        return request;
+    }
+
+    private OCSPResp sendPost(byte[] request) throws IOException {
```
v3rm0n: Can you do this using a RestTemplate instead? Looks like it is really hard to test

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            log.info("Generating and sending OCSPRequest");
+            OCSPResp response = getOCSPResponse(serialNumber);
+
+            log.info("Validating OCSPResponse");
+            checkedResponse = validateOCSPResponse(response);
+
+            log.info("OCSPResponse validated");
+        } catch (OCSPException | IOException e) {
+            log.warn("Couldn't validate serial number");
```
ErkoRisthein: Do we need this `log.warn()` statement? We throw an exception anyway.

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+
+
+    private String validateOCSPResponse(OCSPResp response) throws OCSPException {
+
+        String status = "UNABLE_TO_TEST_USER_CERTIFICATE";
+        switch (response.getStatus()) {
+            case 0:
+                BasicOCSPResp ocspResponseData = (BasicOCSPResp) response.getResponseObject();
+                SingleResp[] responses = ocspResponseData.getResponses();
+                for (SingleResp response1 : responses) {
```
ErkoRisthein: Would be better to rename `response1` to just `response`

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        } catch ( IOException e) {
+            throw new AuthenticationException(AuthenticationException.Code.INTERNAL_ERROR, "Unable to aquire resource", e);
+        }
+
+    }
+
+    public URI getResponderURI(X509Certificate certificate) {
+        AuthorityInfoAccessExtension extension = ((X509CertImpl) certificate).getAuthorityInfoAccessExtension();
+        List<AccessDescription> accessDescriptions = extension.getAccessDescriptions();
+        AccessDescription result = accessDescriptions.stream().filter(accessDescription -> accessDescription.getAccessMethod().equals(AccessDescription.Ad_OCSP_Id)).findFirst().orElse(null);
```
v3rm0n: Instead of orElse(null) can be orElseThrow

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
...
+        OCSPResp response = sendPost(request.getEncoded());
+        return response;
+    }
+
+
+    private String validateOCSPResponse(OCSPResp response) throws OCSPException {
+
+        String status = "UNABLE_TO_TEST_USER_CERTIFICATE";
+        switch (response.getStatus()) {
+            case 0:
```
ErkoRisthein: Magic number. Can we extract well-named constants for the `case 0` and `case 1 ` to make it easier to understand? I.e. what does response status 0 or 1 mean.

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPUtils.java
```diff
@@ -0,0 +1,84 @@
...
+import java.net.URLConnection;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.CertificateParsingException;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+import java.util.List;
+
+@Component
+public class OCSPUtils {
```
v3rm0n: Missing tests

src/main/java/ee/tuleva/onboarding/auth/ocsp/OnlineCertificateStatusProtocolService.java
```diff
@@ -0,0 +1,64 @@
...
+import java.net.URI;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.X509Certificate;
+
+import static com.codeborne.security.AuthenticationException.Code.valueOf;
+
+@Service
+@Slf4j
+@AllArgsConstructor
+public class OnlineCertificateStatusProtocolService {
```
v3rm0n: Missing tests

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
```diff
@@ -0,0 +1,155 @@
+package ee.tuleva.onboarding.auth.ocsp;
+
+import com.codeborne.security.AuthenticationException;
```
ErkoRisthein: Maybe we should define our own `AuthenticationException` class instead. Otherwise we're dependent on the `mobile-id` library.

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPUtils.java
```diff
@@ -0,0 +1,84 @@
...
+    public static final String BEGIN_CERT = "-----BEGIN CERTIFICATE-----";
+    public static final String END_CERT = "-----END CERTIFICATE-----";
+    public final static String LINE_SEPARATOR = System.getProperty("line.separator");
+
+
+    public String getIssuerCertificate(X509Certificate certificate) {
```
ErkoRisthein: Can we make this method shorter? I.e. extract some private methods.

docker-compose.yml
```diff
@@ -1,9 +1,10 @@
 version: '2'
 services:
   database:
-    image: postgres
+    image: postgres:9.6.12
```
ErkoRisthein: Why are we using such an old version of Postgres?
The current version is 12.2
9.6 originally came out in 2016

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPRequestSpec.groovy
```diff
@@ -0,0 +1,183 @@
+package ee.tuleva.onboarding.auth.ocsp
+
+
+import spock.lang.Specification
+
+import java.security.cert.X509Certificate
+
+class OCSPRequestSpec extends Specification {
+
+    def "Test if certificate is valid for (TEST of ESTEID2018.pem.cer)"() {
```
ErkoRisthein: Are these tests running against the real OCSP service? We should mock it. 
Tests should not be dependent on external resources.

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPRequestSpec.groovy
```diff
@@ -0,0 +1,183 @@
...
+        given:
+        OCSPRequest request = new OCSPRequest(caToCheckValid, "http://aia.demo.sk.ee/ee-govca2018")
+        def serialNumberToCheck = new BigInteger("71907861382765673730662460475615088197")
+        def expectedResponse = "GOOD"
```
ErkoRisthein: Maybe the method should return a Response enum instead of a String?

ErkoRisthein:
Conceptually, the `OCSPRequest` does multiple things:
- it _is_ the Request entity
- it _sends_ the request
I think these should be 2 separate classes - i.e. an `OCSPService` class (as a Spring bean) and an `OCSPRequest` that holds the CA to check.

ErkoRisthein:
Code coverage is problematic. You should always aim to increase the coverage, if possible.
Currently the master branch coverage is 83% -> your patch has only 50% coverage.

docker-compose.yml
```diff
@@ -1,9 +1,10 @@
...
     ports:
       - "5432:5432"
     environment:
       - POSTGRES_USER=tuleva-onboarding
       - POSTGRES_DB=tuleva-onboarding
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
ErkoRisthein: I think specifying a `POSTGRES_PASSWORD` and changing the jdbc URL would be a better solution.

docker-compose.yml
v3rm0n: yeah, but this is the version in AWS, we should use the same in Docker. Until we upgrade the DB

src/main/java/ee/tuleva/onboarding/auth/exception/AuthenticationException.java
```diff
@@ -0,0 +1,40 @@
+package ee.tuleva.onboarding.auth.exception;
+
+import lombok.Getter;
+
+public class AuthenticationException extends RuntimeException {
+  public enum Code {
+    INVALID_INPUT("Sisendparameetrid mittekorrektsel kujul"),
```
ErkoRisthein: Can we have these error messages in English instead?

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPFixture.java
```diff
@@ -0,0 +1,174 @@
...
+import sun.security.x509.CertificateX509Key;
+import sun.security.x509.GeneralName;
+import sun.security.x509.URIName;
+import sun.security.x509.X500Name;
+import sun.security.x509.X509CertImpl;
+import sun.security.x509.X509CertInfo;
+
+public class OCSPFixture {
+  public static String sampleExampleServer = "http://aia.sk.ee/esteid2015";
+  public static String getSampleCertificatePEM =
```
ErkoRisthein: Maybe we should save these as files and read them from the resources folder?

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPFixture.java
```diff
@@ -0,0 +1,174 @@
...
+
+  public static String certToString(X509Certificate cert) {
+    StringWriter sw = new StringWriter();
+    try {
+      sw.write("-----BEGIN CERTIFICATE-----\n");
+      sw.write(
+          DatatypeConverter.printBase64Binary(cert.getEncoded()).replaceAll("(.{64})", "$1\n"));
+      sw.write("\n-----END CERTIFICATE-----\n");
+    } catch (CertificateEncodingException e) {
+      e.printStackTrace();
```
ErkoRisthein: Avoid printStackTrace(); use a logger call instead

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPServiceSpec.groovy
```diff
@@ -0,0 +1,122 @@
...
+import org.bouncycastle.cert.ocsp.OCSPRespBuilder
+import org.springframework.http.HttpMethod
+import org.springframework.http.ResponseEntity
+import org.springframework.web.client.RestTemplate
+import spock.lang.Specification
+
+import static org.springframework.http.HttpStatus.OK
+
+class OCSPServiceSpec extends Specification {
+    RestTemplate restTemplate = Mock(RestTemplate);
```
ErkoRisthein: Use a `TestRestTemplate` instead of mocking it: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/web/client/TestRestTemplate.html

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPUtilsSpec.groovy
```diff
@@ -0,0 +1,80 @@
+package ee.tuleva.onboarding.auth.ocsp
+
+import ee.tuleva.onboarding.auth.exception.AuthenticationException
+import spock.lang.Specification
+
+class OCSPUtilsSpec extends Specification {
+    OCSPUtils utils = new OCSPUtils();
```
ErkoRisthein: No need to use semicolons in groovy :p

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPUtils.java
```diff
@@ -0,0 +1,119 @@
...
+    AuthorityInfoAccessExtension extension =
+        ((X509CertImpl) certificate).getAuthorityInfoAccessExtension();
+    List<AccessDescription> accessDescriptions = extension.getAccessDescriptions();
+    AccessDescription result =
+        accessDescriptions.stream()
+            .filter(
+                accessDescription ->
+                    accessDescription.getAccessMethod().equals(AccessDescription.Ad_OCSP_Id))
+            .findFirst()
+            .orElse(null);
```
ErkoRisthein: You can just do `.orElseThrow(() -> new AuthenticationException(...))`

src/test/groovy/ee/tuleva/onboarding/OnboardingServiceApplicationSpec.groovy
```diff
@@ -7,12 +7,21 @@ import spock.lang.Specification
...
-
     @Autowired
     WebApplicationContext context
 
     def "context loads"() {
         expect:
         context != null
     }
+
+    def "test"() {
```
ErkoRisthein: This doesn't really test anything

ristosaar: I'm not really sure what happened here. This wasn't supposed to be committed :sweat_smile:

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPRequest.java
v3rm0n: https://github.com/TulevaEE/onboarding-service/pull/215

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPService.java
```diff
@@ -0,0 +1,125 @@
...
+  public OCSPResponseType checkCertificate(OCSPRequest request) {
+    X509Certificate certificate = request.getCertificate();
+    if (hasCertificateExpired(certificate)) {
+      return EXPIRED;
+    }
+    return checkCertificateStatus(request);
+  }
+
+  private boolean hasCertificateExpired(X509Certificate certificate) {
+    Clock clock = Clock.systemDefaultZone();
```
v3rm0n: This is not better than using System.currentTimeMillis. You can still not mock it. Dependency injection is the keyword here. Create a Clock bean and inject it.

ErkoRisthein: Take a look at how it's injected and mocked in: https://github.com/TulevaEE/onboarding-service/blob/master/src/test/groovy/ee/tuleva/onboarding/conversion/UserConversionServiceSpec.groovy#L35

src/test/groovy/ee/tuleva/onboarding/auth/ocsp/OCSPServiceSpec.groovy
```diff
@@ -11,20 +12,29 @@ import org.bouncycastle.cert.ocsp.OCSPReqBuilder
...
+
+@RestClientTest(OCSPService)
+@ContextConfiguration(classes = [ClockConfig.class, OCSPService.class])
 class OCSPServiceSpec extends Specification {
     @Autowired
-    TestRestTemplate restTemplate
+    private final MockRestServiceServer server
     @Autowired
-    OCSPService service;
+    private final OCSPService service
```
ErkoRisthein: Fields can't be final if you set them with `@Autowired`

ErkoRisthein:
There is not happy case test for the `OCSPService.validateOCSPResponse()` method.
I.e. where status = GOOD or REVOKED.
Other than that, looks good 🍺

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPService.java
```diff
@@ -0,0 +1,130 @@
...
+  private final RestOperations restTemplate;
+  private final Clock clock;
+
+  public OCSPService(RestTemplateBuilder restTemplateBuilder, Clock clock) {
+    restTemplate =
+        restTemplateBuilder.setConnectTimeout(ofSeconds(60)).setReadTimeout(ofSeconds(60)).build();
+    this.clock = clock;
+  }
+
+  public OCSPResponseType checkCertificate(OCSPRequest request) {
```
ErkoRisthein: Maybe this should be the only public method of the class?

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPService.java
```diff
@@ -0,0 +1,130 @@
...
+      log.info("Generating and sending OCSPRequest");
+      OCSPResp response = getOCSPResponse(request);
+      return validateOCSPResponse(response);
+    } catch (OCSPException | RestClientException | IOException e) {
+      throw new AuthenticationException(
+          UNABLE_TO_TEST_USER_CERTIFICATE, "Couldn't validate serial number", e);
+    }
+  }
+
+  public OCSPResponseType validateOCSPResponse(OCSPResp response) throws OCSPException {
```
ErkoRisthein: Should be private

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPService.java
```diff
@@ -0,0 +1,130 @@
...
+    byte[] derCert = response.getBody();
+    final Base64.Encoder encoder = Base64.getMimeEncoder(64, LINE_SEPARATOR.getBytes());
+    final String encodedCertText = new String(encoder.encode(derCert));
+    final String pemCert =
+        BEGIN_CERT + LINE_SEPARATOR + encodedCertText + LINE_SEPARATOR + END_CERT;
+
+    return pemCert;
+  }
+
+  public OCSPResponseType checkCertificateStatus(OCSPRequest request) {
```
ErkoRisthein: Should be private

src/main/java/ee/tuleva/onboarding/auth/ocsp/OCSPAuthService.java
```diff
@@ -0,0 +1,70 @@
...
+        ocspUtils.generateOCSPRequest(cert, certificationAuthority, responderURI.toString());
+
+    OCSPResponseType result = service.checkCertificate(request);
+    validateGoodResult(result);
+
+    return getCreateCertificateResponse(cert);
+  }
+
+  @NotNull
+  CheckCertificateResponse getCreateCertificateResponse(X509Certificate cert) {
```
ErkoRisthein: Should be private

PR #231, opened 2020-04-22
New smart id signing

ErkoRisthein:
- Removed the dependency on the deprecated DigiDocService
- BDOC container now generated using the DigiDoc4j library
- Removes the Async executor that was mutating the HTTPSession
- No blocking threads, all polling done on the client side


src/main/java/ee/tuleva/onboarding/mandate/MandateController.java
```diff
@@ -77,33 +74,33 @@ public MobileSignatureResponse startMobileIdSignature(
...
+        sessionStore.get(MobileIdSignatureSession.class);
     MobileIdSignatureSession session =
         signatureSession.orElseThrow(IdSessionException::mobileSignatureSessionNotFound);
 
     String statusCode =
         mandateService.finalizeMobileIdSignature(
             authenticatedPerson.getUserId(), mandateId, session);
 
-    return new MandateSignatureStatusResponse(statusCode);
+    return new MobileSignatureStatusResponse(statusCode, session.challenge);
```
ristosaar, line 103: I know it's from a library but is there a reason there is no getter for this field?

ErkoRisthein: There is. The `@Data` annotation adds getters.

src/main/java/ee/tuleva/onboarding/config/DigiDocConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+public class DigiDocConfiguration {
+
+    @Bean
+    @Profile("production")
+    public org.digidoc4j.Configuration digiDocConfigProd() {
+        return new org.digidoc4j.Configuration(org.digidoc4j.Configuration.Mode.PROD);
+    }
+
+    @Bean
+    @Profile("!production")
```
v3rm0n, line 17: If you have this kind of conditional beans then it is usually better for the negative case to use `@ConditionalOnMissingBean(Configuration.class)` instead of `@Profile~ makes the configuration a bit more resilient. Random example: current case would break if you add another Profile on top of the other bean and there would be two beans in the application context.

Even better would be configuring it in properties `@ConditionalOnProperty(prefix="digidoc.config", name="type", havingValue="prod")` then you can turn it on and off without switching the entire profile.

PR #235, opened 2020-05-08
Feature/email on confirm

ErkoRisthein:


src/main/java/ee/tuleva/onboarding/user/User.java
```diff
@@ -80,4 +80,15 @@ protected void onCreate() {
     protected void onUpdate() {
         updatedDate = Instant.now();
     }
+
+    public String getNameSuffix() {
```
ErkoRisthein, line 84: This is mandate-file specific logic. Not strictly related to the User domain.

src/test/groovy/ee/tuleva/onboarding/mandate/listener/MandateEmailSenderSpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+    @SpringBean
+    MandateEmailService mandateEmailService = Mock(MandateEmailService)
+
+    def "send email when second pillar mandate event was received" () {
+        given:
+        User user = sampleUser().build()
+        SecondPillarMandateCreatedEvent event = new SecondPillarMandateCreatedEvent(user, 123, "123".bytes)
+        when:
+        publisher.publishEvent(event)
+        TimeUnit.SECONDS.sleep(1)
```
ErkoRisthein, line 34: Why do we need to sleep? Bad practice to do it in tests.

PR #247, opened 2020-05-27
Bug/holding detail job

songxunzhao:


ErkoRisthein:
Debug level logging is off in production. Use `log.info()` instead.

PR #250, opened 2020-06-25
New MARKET index logic

aleksandraisabel:
New Global Stock index list needed to be created with MARKET and GLOBAL STOCK indices (change on Jan 01 2020).

Existing lists are pulled with SQL query, with appropriate alterations values are added to a new created list called NEW_GLOBAL_STOCK_INDEX.

Some exceptions and code errors might be found as database connection was created but values not portrayed.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
+++ b/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
@@ -2,23 +2,27 @@ package ee.tuleva.onboarding.comparisons.fundvalue.retrieval;
 
 import ee.tuleva.onboarding.comparisons.fundvalue.FundValue;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
 import org.springframework.stereotype.Service;
 
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.sql.*;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.time.LocalDate;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
 
 import static java.util.stream.Collectors.toList;
 
 @Slf4j
 @Service
-public class GlobalStockIndexCreator implements ComparisonIndexRetriever  {
+public class GlobalStockIndexCreator implements ComparisonIndexRetriever {
     public static final String KEY = "NEW_GLOBAL_STOCK_INDEX";
 
-    //private NamedParameterJdbcTemplate jdbcTemplate;
+    NamedParameterJdbcTemplate jdbcTemplate;
 
     private static final String SQL = "" +
         "SELECT * " +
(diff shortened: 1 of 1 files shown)


src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+@Slf4j
+@Service
+public class GlobalStockIndexCreator implements ComparisonIndexRetriever {
+    public static final String KEY = "NEW_GLOBAL_STOCK_INDEX";
+
+    NamedParameterJdbcTemplate jdbcTemplate;
```
ErkoRisthein, line 25: Separation of concerns. Use the Repository design pattern to encapsulate & separate the data access logic from other higher-abstraction-level business logic in this class. 

https://martinfowler.com/eaaCatalog/repository.html

ErkoRisthein:
Name your branches by the feature name. I.e. instead of `isabel-branch` name it `feature/new-market-index` or something along those lines.

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
```diff
@@ -0,0 +1,95 @@
...
+    private List<FundValue> getStockValues() {
+        try {
+            return createStockList();
+        } catch (IOException e) {
+            throw new IllegalStateException("Could not get Global Stock Index values", e);
+        }
+    }
+
+
+    private List<FundValue> createStockList() throws IOException {
```
ErkoRisthein, line 65: `IOException` is never thrown from this method.

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
```diff
@@ -0,0 +1,95 @@
...
+        return stockValues.stream().filter(fundValue -> {
+            LocalDate date = fundValue.getDate();
+            return (startDate.isBefore(date) || startDate.equals(date)) && (endDate.isAfter(date) || endDate.equals(date));
+        }).collect(toList());
+    }
+
+    private List<FundValue> getStockValues() {
+        try {
+            return createStockList();
+        } catch (IOException e) {
```
ErkoRisthein, line 59: There's nothing to catch here since an `IOException` is never thrown from that method.

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
ErkoRisthein, line 65: `createStockList` is not a good method name. Generally speaking, you should never use the name of the data structure (i.e. `List` in this case) in method/variable names. `createStocks` would be better since it gives you flexibility to change the underlying data structure without risking accidentally contradicting (or deceiving) your readers. I.e. if you happen to change the return type of the method to a `Map` in the future, but forget to change the name of the method then you are effectively lying about the return type of your method (i.e. a `createList()` method actually creates a `Map`). So to avoid this kind of human errors in the future, it's good practice to never expose the return type in your method names (or data types in variable names).

Also a mandatory read:
Clean Code Chapter 2 - Writing Meaningful Names

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
ErkoRisthein: In other words, this class should depend on the `FundValueRepository` (which, in turn, depends on the `jdbcTemplate`) not directly on the `jdbcTemplate`.

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
```diff
@@ -0,0 +1,95 @@
...
+import java.sql.SQLException;
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.stream.Collectors.toList;
+
+@Slf4j
+@Service
+public class GlobalStockIndexCreator implements ComparisonIndexRetriever {
```
ErkoRisthein, line 22: If this class implements the `ComparisonIndexRetriever` interface then it IS-A `Retriever`. Therefore, it should be named accordingly - i.e. `GlobalStockIndexRetriever` or something along those lines.

https://stackoverflow.com/questions/36162714/what-is-the-difference-between-is-a-relationship-and-has-a-relationship-in-java

ErkoRisthein:
Missing unit and integration tests. After making the changes according to the comments, you should have integration tests for the `FundValueRepository` and unit tests for the `GlobalStockIndexRetriever`.

src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/GlobalStockIndexCreator.java
```diff
@@ -0,0 +1,95 @@
...
+        for (int i = 0; i < jdbcValues.size(); i++) {
+            FundValue fundvalue = jdbcValues.get(i);
+
+            LocalDate date = fundvalue.getDate();
+            LocalDate changeDate = LocalDate.of(2019, 12, 31);
+            BigDecimal value = fundvalue.getValue();
+
+            if (date.isAfter(changeDate)) {
+                Double convertedNumber = value.doubleValue();
+                convertedNumber = convertedNumber / 13.3883094;
```
ErkoRisthein, line 78: `13.3883094` is a magic number. Better to extract into a variable (or a constant) and give it a meaningful name so the readers of your code can understand what this number is.

https://stackoverflow.com/a/47902/2102820

PR #257, opened 2020-07-08
Index change in API level

aleksandraisabel:
WorldIndexValueRetriever.KEY changed to new index key, so that UNION_STOCK_INDEX could be portrayed in pension account comparison box. 

Tests modified so that they run and check validity of new generated index.

Unit tests added to UnionStockIndexRetrieverSpec.


src/main/java/ee/tuleva/onboarding/comparisons/returns/provider/IndexReturnProvider.java
```diff
@@ -27,7 +27,7 @@
 
     private static final String EPI = EPIFundValueRetriever.KEY;
 
-    private static final String MARKET = WorldIndexValueRetriever.KEY;
+    private static final String MARKET = UnionStockIndexRetriever.KEY;
```
ErkoRisthein, line 30: 2 problems with this:
1. It's misleading.
Previously, it was `String MARKET = "MARKET"` in short. (1:1 match between the API name and Retriever KEY)
Now, it's `String MARKET = "UNION_STOCK_INDEX"` in short. (mismatch between API name and Retriever KEY)
2. It actually changes the API. So you would need to change your clients as well. 
I.e. you cannot do `GET /returns?from=2003-01-07&keys[]=MARKET` anymore. 
You have to do `GET /returns?from=2003-01-07&keys[]=UNION_STOCK_INDEX`. 
This makes the API change backwards incompatible, which is another topic up for debate - whether we want to keep it backwards compatible or not right now.

I propose we stick to the 1:1 mapping between the API names and Retriever KEYs - and go forward with the backwards incompatible change. Why? It just keeps things simple and avoids confusion. 

The alternative would be to have "MARKET" in the API layer and "UNION_STOCK_INDEX" in the service/database layer. This is confusing since we also have a "MARKET" time series in the database.
In short, the mapping would be counter-intuitively MARKET=UNION_STOCK_INDEX not MARKET=MARKET.

src/test/groovy/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/UnionStockIndexRetrieverSpec.groovy
```diff
@@ -8,10 +8,45 @@ import java.time.LocalDate
...
+
+    def "it is configured for the right fund"() {
+        when:
+        def retrievalFund = retriever.getKey()
+
+        then:
+        retrievalFund == UnionStockIndexRetriever.KEY
+    }
+
+    def "it should work with empty data files"() {
```
ErkoRisthein, line 27: should work with empty data files?

src/main/java/ee/tuleva/onboarding/comparisons/returns/provider/IndexReturnProvider.java
ErkoRisthein: Rename `String MARKET` to `String UNION_STOCK_INDEX`.

ErkoRisthein:
- Resolve merge conflicts.
- Create another PR for the onboarding-client.
- Merge them both at the same time (so the service & client would be released together since their API needs to match with each other).

src/main/java/ee/tuleva/onboarding/comparisons/returns/provider/IndexReturnProvider.java
aleksandraisabel: String MARKET changed to String UNION_STOCK_INDEX

src/test/groovy/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/UnionStockIndexRetrieverSpec.groovy
aleksandraisabel: empty data files changed to missing index values

PR #260, opened 2020-07-30
Delete mailchimp and its dependencies

aleksandraisabel:
Delete mailchimp and its dependencies. Checked epis-service and onboarding-client and verified no dependencies alongside.


ErkoRisthein:
It's probably safe to delete the `com.ecwid:maleorang` gradle dependency as well.

PR #321, opened 2021-01-12
Add extra checks to id-card certificate

ErkoRisthein:


etc/eb/.ebextensions/nginx/conf.d/01_ssl_proxy.conf
```diff
@@ -121,6 +129,7 @@ server {
 
     location /idLogin {
         expires -1;
+        # proxy_pass http://192.168.65.2:9000;
         proxy_pass http://localhost:5000;
```
v3rm0n, line 133: Use $backend here as well.

etc/eb/.ebextensions/nginx/conf.d/01_ssl_proxy.conf
```diff
@@ -1,3 +1,11 @@
+map $host $frontend {
+  default "http://pension.tuleva.ee.s3-website.eu-central-1.amazonaws.com";
+  # default "http://192.168.65.2:3000";
```
v3rm0n, line 3: What is this IP? If local then use `127.0.0.1` instead. Maybe you can use this trick in the other nginx file which includes this file to have it as default https://stackoverflow.com/a/30083902

README.md
```diff
@@ -110,6 +110,24 @@ curl --location --request POST 'https://pension.tuleva.ee/api/oauth/token' \
...
+
+In order to test ID-card locally, you need to run nginx locally with the right certificates and the right domain names.
+
+1. Add tuleva certs to `./nginx` (4 files)
+2. Update ```$frontend``` and `$backend` urls in `etc/eb/.ebextensions/nginx/conf.d/01_ssl_proxy.conf`
+3. Add to `hosts` file:
+   ```
+   127.0.0.1 id.tuleva.ee
+   127.0.0.1 pension.tuleva.ee
+   127.0.0.1 onboarding-service.tuleva.ee
```
v3rm0n, line 123: Should we get rid of this domain and only use pension? :thinking:

v3rm0n (approved):

etc/eb/.ebextensions/nginx/conf.d/01_ssl_proxy.conf
ErkoRisthein: Docker localhost

PR #346, opened 2021-03-11
Add pending withdrawal info to conversion API

v3rm0n:
- Return pendingWithdrawal as true when there are PENDING status
applications of type WITHDRAWAL or EARLY_WITHDRAWAL

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/mandate/application/Application.java
+++ b/src/main/java/ee/tuleva/onboarding/mandate/application/Application.java
@@ -1,5 +1,6 @@
 package ee.tuleva.onboarding.mandate.application;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import ee.tuleva.onboarding.epis.mandate.ApplicationStatus;
 import java.time.Instant;
 import lombok.Builder;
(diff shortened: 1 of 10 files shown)


src/main/java/ee/tuleva/onboarding/mandate/application/Application.java
```diff
@@ -17,6 +18,16 @@
   private ApplicationStatus status;
   private ApplicationDetails details;
 
+  @JsonIgnore
+  public boolean isPending() {
+    return status.equals(ApplicationStatus.PENDING);
+  }
+
+  @JsonIgnore
```
ErkoRisthein, line 26: A good reason to extract an `ApplicationDto` - so we wouldn't accidentally leak this kind of methods thru the REST API.

src/main/java/ee/tuleva/onboarding/mandate/application/ApplicationService.java
```diff
@@ -24,10 +21,14 @@
...
-  private final MandateCancellationService mandateCancellationService;
   private final LocaleService localeService;
   private final FundRepository fundRepository;
 
+  public boolean hasPendingWithdrawals(Person person) {
+    return get(person).stream()
+        .anyMatch(application -> application.isPending() && application.isWithdrawal());
+  }
+
   public List<Application> get(Person person) {
```
ErkoRisthein, line 32: `getApplications()` would be a better name

ErkoRisthein (approved):

PR #385, opened 2021-08-02
Add zero funds with subtractions

MatsJohanson:
This change should allow users who are confused about their fund finance movements to remember that they had non-tuleva funds active in the past.

Solves https://app.asana.com/0/229163390356948/1199360837555677

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/account/AccountStatementService.java
+++ b/src/main/java/ee/tuleva/onboarding/account/AccountStatementService.java
@@ -7,7 +7,9 @@ import static org.apache.commons.lang3.ObjectUtils.compare;
 import ee.tuleva.onboarding.auth.principal.Person;
 import ee.tuleva.onboarding.epis.EpisService;
 import ee.tuleva.onboarding.epis.account.FundBalanceDto;
+
 import java.util.List;
+
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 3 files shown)


src/test/groovy/ee/tuleva/onboarding/account/AccountStatementServiceSpec.groovy
```diff
@@ -57,8 +56,12 @@ class AccountStatementServiceSpec extends Specification {
     def activeNonZeroFund = FundBalanceDto.builder().isin("4").value(ONE).activeContributions(true).build()
 
     episService.getAccountStatement(person) >> [nonActiveZeroFund, nonActiveNonZeroFund, activeZeroFund, activeNonZeroFund]
-    fundBalanceConverter.convert(_, person) >> { FundBalanceDto fundBalanceDto, _ ->
-      FundBalance.builder().fund(Fund.builder().isin(fundBalanceDto.isin).build()).build()
+    fundBalanceConverter.convert(_, _) >> { FundBalanceDto fundBalanceDto, _ ->
+      FundBalance.builder()
+        .fund(Fund.builder().isin(fundBalanceDto.isin).build())
```
labidas, line 61: Not the cleanest way to mock out the converter behaviour here, but will improve in the future.

ErkoRisthein:
I think we should think about how to consider payouts only, not all subtractions. Subtractions are both payouts and transfers between funds. We want to ignore transfers, and only consider payouts.

MatsJohanson:
> I think we should think about how to consider payouts only, not all subtractions. Subtractions are both payouts and transfers between funds. We want to ignore transfers, and only consider payouts.

Had a product discussion with Tõnu yesterday and got some clarity. We want to show all 3rd pillar funds that have had activity since they all affect the displayed percentages which caused questions.

I changed the PR so we only filter out zero sum inactive second pillars.

src/main/java/ee/tuleva/onboarding/account/AccountStatementService.java
```diff
@@ -25,14 +25,17 @@
...
-        .filter(
-            fundBalanceDto ->
-                compare(ZERO, fundBalanceDto.getValue()) != 0
-                    || fundBalanceDto.isActiveContributions())
+        .filter(this::secondPillarNotEmpty)
         .map(fundBalanceDto -> convertToFundBalance(fundBalanceDto, person))
         .collect(toList());
   }
 
+  private boolean secondPillarNotEmpty(FundBalanceDto fundBalanceDto) {
```
ErkoRisthein, line 33: `boolean` function names should start with `is...` - i.e. `isSecondPillarNotEmpty` in this case.

ErkoRisthein (approved):

ErkoRisthein:
I think the frontend also filters out founds with 0 amount. The API shows the right results for me, but nothing shows up in the front end.

ErkoRisthein:
e.g. https://github.com/TulevaEE/onboarding-client/blob/11a834a0d1df1c7303f88dba106a418373684515/src/components/thirdPillar/reducer.js#L49
And maybe in some other places as well.

MatsJohanson:
Ah, I tested it on a 2nd pillar fund, but frontend has custom 3rd pillar logic built in. Davai, will improve my input data and ship it in a new PR.

PR #403, opened 2021-09-16
Fix mobile id signing

MatsJohanson:
Currently our mandate signing with mobile id fails, since session tokens are lost between requests. Since we don't validate the login session before initializing Smart-ID and ID-card signature session, it did not affect users using those flows.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,15 +7,10 @@ indent_size = 2
 trim_trailing_whitespace = true
 
 [*.{gradle,gradle.kts}]
-charset = utf-8
-indent_style = space
 indent_size = 4
 insert_final_newline = true
 max_line_length = 120
 
 [*.{java,groovy}]
-charset = utf-8
-indent_style = space
-indent_size = 2
 insert_final_newline = false
 max_line_length = 100
--- a/src/main/java/ee/tuleva/onboarding/mandate/signature/SignatureFile.java
+++ b/src/main/java/ee/tuleva/onboarding/mandate/signature/SignatureFile.java
@@ -1,5 +1,6 @@
 package ee.tuleva.onboarding.mandate.signature;
 
+import java.io.Serial;
 import java.io.Serializable;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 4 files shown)


MatsJohanson:
@ErkoRisthein - do you remember why you reverted the db session change? Would like to be sure it didn't mess anything serious up.

ErkoRisthein:
Something significant broke in production. I don't remember the details, unfortunately.
I guess we can deploy it, test it quickly and immediately roll back.

.editorconfig
```diff
@@ -7,15 +7,10 @@ indent_size = 2
...
-charset = utf-8
-indent_style = space
 indent_size = 4
 insert_final_newline = true
 max_line_length = 120
 
 [*.{java,groovy}]
-charset = utf-8
-indent_style = space
-indent_size = 2
```
ErkoRisthein, line 19: Why remove these?

src/main/java/ee/tuleva/onboarding/mandate/signature/SignatureFile.java
```diff
@@ -1,11 +1,12 @@
 package ee.tuleva.onboarding.mandate.signature;
 
+import java.io.Serializable;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
 @Getter
-public class SignatureFile {
+public class SignatureFile implements Serializable {
```
ErkoRisthein, line 9: Missing `serialVersionUID`

MatsJohanson:
> Something significant broke in production. I don't remember the details, unfortunately.
> I guess we can deploy it, test it quickly and immediately roll back.
> 
> EDIT: Apparently it broke digital signing altogether. Check Slack #tech channel history. Jump to 22 Apr 2020 or search for "Revert session sharing" to see the messages that came before that.

It broke smart-id signing since SignatureFile was not serializable. I tested it and it should work now.

.editorconfig
MatsJohanson: They are defined above with the same values. Just OCD things.

ErkoRisthein (approved):
Cool, let's ship it. 🚀 
Make sure you test the _full flow_ (from login to digital signature) after deployment. 🕶️
I've made that mistake multiple times before -> assuming everything works after a successful login. Don't be me.

PR #408, opened 2021-09-24
Test e-mail changes with snapshotting instead of manual asserts

MatsJohanson:


ErkoRisthein (commented):
I think this is a good utilization of snapshot testing. Better than the tests we had before. Good job.

PR #407, opened 2021-09-23
Separate third pillar mandate e-mail into two parts

MatsJohanson:
In the first e-mail, we're sending out payment details for the third pillar and giving some tips on how to DCA your contributions over time.

The second e-mail will be sent 3 days later to users who have not joined our second pillar, reminding them why our fund is a good option.

PR #412, opened 2021-09-27
Two part 3rd pillar email code review follow ups

MatsJohanson:
Responding to comments from this PR: https://github.com/TulevaEE/onboarding-service/pull/407


src/main/java/ee/tuleva/onboarding/mandate/email/MandateEmailSender.java
```diff
@@ -20,18 +19,11 @@
   @EventListener
   public void sendEmail(AfterMandateSignedEvent event) {
     UserPreferences contactDetails = episService.getContactDetails(event.getUser());
-    PillarSuggestion pillarSuggestion = buildPillarSuggestion(event, contactDetails);
+    ConversionResponse conversion = conversionService.getConversion(event.getUser());
+    int suggestedPillar = event.getPillar() == 2 ? 3 : 2;
```
ErkoRisthein, line 23: Maybe this logic should be encapsulated in the `PillarSuggestion`? I.e. you pass `pillar` as the constructor argument and the `suggestedPillar` is decided inside of the `PillarSuggestion` constructor.

MatsJohanson: > Maybe this logic should be encapsulated in the `PillarSuggestion`? I.e. you pass `pillar` as the constructor argument and the `suggestedPillar` is decided inside of the `PillarSuggestion` constructor.

Also considered it, but thought it might be confusing since we don't have named params in java. But since you're in the same boat, I'll go that route, cheers.

src/main/java/ee/tuleva/onboarding/mandate/email/MandateEmailService.java
```diff
@@ -26,41 +26,50 @@
...
-            "Pensionifondi avaldus",
+            subject,
             getSecondPillarContent(user, mandate, pillarSuggestion, locale),
-            getMandateTags(pillarSuggestion),
+            getSecondPillarMandateTags(pillarSuggestion),
             getMandateAttachments(mandate.getSignedFile(), user, mandate.getId()));
     emailService.send(user, mandrillMessage);
   }
 
+  private List<String> getSecondPillarMandateTags(PillarSuggestion pillarSuggestion) {
```
ErkoRisthein, line 57: Why no tags for the 3rd pillar mandate?

src/main/java/ee/tuleva/onboarding/mandate/event/AfterMandateSignedEvent.java
```diff
@@ -21,15 +21,8 @@ public AfterMandateSignedEvent(Object source, User user, Mandate mandate, Locale
...
-  public static AfterMandateSignedEvent newInstance(
-      Object source, User user, Mandate mandate, Locale locale) {
-    if (mandate.getPillar() == 2) {
-      return new SecondPillarAfterMandateSignedEvent(source, user, mandate, locale);
-    } else if (mandate.getPillar() == 3) {
-      return new ThirdPillarAfterMandateSignedEvent(source, user, mandate, locale);
-    } else {
-      throw new IllegalArgumentException("Event for the pillar type is not available");
-    }
+  public Integer getPillar() {
```
ErkoRisthein, line 24: We could add a `getSuggestedPillar()` method here that does the logic of:
`return mandate.getPillar() == 2 ? 3 : 2;`

src/main/java/ee/tuleva/onboarding/mandate/email/MandateEmailService.java
MatsJohanson: They do have tags, but since they're now in separate methods, they didn't need a method to build them.

src/main/java/ee/tuleva/onboarding/mandate/email/MandateEmailSender.java
ErkoRisthein: Or even better, just add a method called `pillarSuggestion.getSuggestedPillar()` that encapsulates that logic

PR #413, opened 2021-09-28
Add temporary endpoint to record mandate confirm events

MatsJohanson:
This is necessary to gather data about the mandate flow usage. In the future, when we have solved user behavior analysis with a data driven tool such as Mixpanel or Heap we'll remove this endpoint.


ErkoRisthein:
Temporary endpoints are usually not temporary. I would rather build the correct solution (i.e. mixpanel events in the frontend).

PR #415, opened 2021-09-30
Add support for fetching PIK transfer applications

MatsJohanson:


src/main/java/ee/tuleva/onboarding/conversion/UserConversionService.java
```diff
@@ -182,11 +183,9 @@ private boolean isTransfersComplete(
...
 
   private boolean isConvertedFundManager(Exchange exchange) {
-    return exchange
-        .getTargetFund()
-        .getFundManager()
-        .getName()
-        .equalsIgnoreCase(CONVERTED_FUND_MANAGER_NAME);
+    FundDto targetFund = exchange.getTargetFund();
+    return targetFund != null
+        && targetFund.getFundManager().getName().equalsIgnoreCase(CONVERTED_FUND_MANAGER_NAME);
```
ErkoRisthein, line 188: According to the [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter#In_object-oriented_programming) we should ask for the fund manager name directly from the targetFund. I.e. refactor it in a way that we would do: `targetFund.getFundManagerName()` instead.

src/main/java/ee/tuleva/onboarding/mandate/application/ApplicationService.java
```diff
@@ -66,7 +68,7 @@ public boolean hasPendingWithdrawals(Person person) {
   }
 
   private List<TransferApplication> groupTransfers(List<ApplicationDTO> transferApplications) {
-    val locale = localeService.getCurrentLocale();
+    String language = localeService.getCurrentLocale().getLanguage();
```
ErkoRisthein, line 71: According to the [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter#In_object-oriented_programming) we should ask for the language directly from the localeService. I.e. refactor it in a way that we would do: `localeService.getCurrentLanguage()` instead.

MatsJohanson: Turns out there already was such a method. I'm a fool!

src/main/java/ee/tuleva/onboarding/conversion/UserConversionService.java
```diff
@@ -182,11 +183,9 @@ private boolean isTransfersComplete(
...
 
   private boolean isConvertedFundManager(Exchange exchange) {
-    return exchange
-        .getTargetFund()
-        .getFundManager()
-        .getName()
-        .equalsIgnoreCase(CONVERTED_FUND_MANAGER_NAME);
+    FundDto targetFund = exchange.getTargetFund();
+    return targetFund != null
+        && targetFund.getFundManagerName().equalsIgnoreCase(CONVERTED_FUND_MANAGER_NAME);
```
ErkoRisthein, line 188: 1 option would be to move this logic into the `FundDto` class as well. I.e. so we could do something like:
`targetFund.isConverted()`

src/main/java/ee/tuleva/onboarding/mandate/application/ApplicationService.java
```diff
@@ -104,6 +105,18 @@ public boolean hasPendingWithdrawals(Person person) {
         .collect(toList());
   }
 
+  private FundDto getTargetFund(MandateFundsTransferExchangeDTO exchangeDTO, String language) {
+    String targetFundIsin = exchangeDTO.getTargetFundIsin();
+    if (targetFundIsin == null) return null;
```
ErkoRisthein, line 110: Why return null? This might propagate into the system and cause a `NullpointerException` from somewhere else.

ErkoRisthein: Is is null when there's a `targetPik` instead?

ErkoRisthein: Would be great to have some validation for this somewhere - i.e. in the `TransferApplicationDetails.Exchange` constructor. So it would quickly throw an exception if something's wrong. We're effectively doing an [XOR](https://www.maximintegrated.com/content/dam/images/glossary/xor-gate-symbol.jpg) on `targetFund` and `targetPik`. *One and only one* should be filled. Not both. Not neither.

src/main/java/ee/tuleva/onboarding/mandate/application/ApplicationService.java
```diff
@@ -104,6 +105,18 @@ public boolean hasPendingWithdrawals(Person person) {
...
+
+    Fund targetFund = fundRepository.findByIsin(targetFundIsin);
+    if (targetFund == null) {
+      throw new IllegalArgumentException(
+          "Fund with ISIN \"" + targetFundIsin + "\" not found in database!");
```
ErkoRisthein, line 115: I like to keep the variables at the end of the log message to keep them easily _greppable_. 

Right now if you would like to find all instances of this exception from Papertrail, you would have to search for `"Fund with ISIN"` -> and it might find other logging statements instead as well - since it's a fairly generic phrase. 
Or search for `"not found in database"` which is also pretty generic and might find other log lines as well.

However, if you would log it like this:
```
throw new IllegalArgumentException("Fund not found in the database: targetFundIsin=" + targetFundIsin);
```
It would be very easy to search for this exact error message from Papertrail by the _specific_ search string:
"Fund not found in the database: targetFundIsin="

I hope I'm making sense with this. The generic logging pattern would be: 
```Error message: var1={var1}, var2={var2}...```

File: src/main/resources/db/migration/V1_51__add_target_pik_to_fund_transfer_exchange.sql [hunk 6]
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE fund_transfer_exchange
+  ADD COLUMN target_pik VARCHAR NULL;
```
ErkoRisthein, line 2: Would be great to have some kind of a limit on this column length. Is PIK an IBAN code? In that case it's up to 34 characters: https://en.wikipedia.org/wiki/International_Bank_Account_Number
Ie. `VARCHAR(34)`

src/test/groovy/ee/tuleva/onboarding/mandate/application/ApplicationServiceSpec.groovy
```diff
@@ -107,6 +110,24 @@ class ApplicationServiceSpec extends Specification {
...
+      type == TRANSFER
+      status == PENDING
+      creationTime == ClockFixture.now
+      with(details) {
+        sourceFund.isin == "AE123232334"
+        exchanges.size() == 1
+        with(exchanges[0]) {
+          sourceFund.isin == "AE123232334"
+          targetFund == null
+          targetPik == "targetPik"
```
ErkoRisthein, line 123: I prefer to use "realistic" data in tests. It gives a better sense of what a PIK really should look like and makes the domain easier to understand for new engineers joining the team. Tests are effectively the _specification_ - and they should be real-life example inputs and real-life example outputs.

 For example, if it's an IBAN code then just generate a random IBAN: http://randomiban.com/?country=Estonia
e.g. `EE801281685311741971`

src/test/groovy/ee/tuleva/onboarding/mandate/application/ApplicationDtoFixture.groovy
```diff
@@ -16,7 +16,23 @@ class ApplicationDtoFixture {
       .id(123L)
       .currency("EUR")
       .sourceFundIsin("source")
-      .fundTransferExchanges([new MandateDto.MandateFundsTransferExchangeDTO("processId", 1.0, "source", "target")])
+      .fundTransferExchanges([
+        new MandateDto.MandateFundsTransferExchangeDTO("processId", 1.0, "source", "target", null)
```
ErkoRisthein, line 20: Always prefer readability above anything else in tests. For example, if it increases readability, use [named parameters](http://docs.groovy-lang.org/latest/html/documentation/#_named_parameters_2) in Groovy. So you could do something like: 
```
new MandateFundsTransferExchangeDTO(processId: "b22ba9", amount: 1.0, sourceFundIsin: "EE1312312313", targetFundIsin: "EE2323232326", targetPik: null)
```

MatsJohanson:
All the comments make sense. I got some time and pulled them into a follow-up PR: https://github.com/TulevaEE/onboarding-service/pull/430/

UserConversionService is starting to look a lot cleaner bit by bit.

PR #429, opened 2021-10-06
Add logging to missing mobile id sessions

MatsJohanson:
We're still getting client reports about not being able to sign documents with mobile id. Let's add some logging to find out how many such situations exist and try to debug.


MatsJohanson:
> * Why remove the messages? It makes the API harder to use & harder to debug/reason about problems. Also, lowers the readability/understandability/self-documentation of the backend code. I have a hard time understanding the purpose of this change.
> * The frontend depends on the messages, is it compatible with this change?
> 
> Good RESTful API error messages are also `human-readable`. https://blog.restcase.com/rest-api-error-handling-problem-details-response/

The issue here was that we were passing on internal error messages (Postgres and EPIS internals) to the customers. They don't understand what to do with them but as per PenTest reports, a malicious actor could use them against us. Now we're just passing a code  which is translated by the front-end. 

There are some missing translation keys, but they will arrive in the next client pull request.

MatsJohanson:
I'll remove the messages commit from this PR to get the logging live. We can continue the discussion about messages later :)

PR #444, opened 2021-10-21
Update Smart-ID library 1.6.1 -> 2.1.1

mats-johanson:


MatsJohanson:
> Looks good, good job muscle What's the patch code coverage?

For some reason, Jacoco fails to generate the report on my machine so I'll fix and add CodeCov back in a separate task.

PR #467, opened 2021-11-09
Allow users with duplicate e-mails to reach contact details page

MatsJohanson:
Currently they just get an error when trying to sign up. This allows them to see what is actually wrong.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/user/UserService.java
+++ b/src/main/java/ee/tuleva/onboarding/user/UserService.java
@@ -50,7 +50,7 @@ public class UserService {
     return save(user);
   }
 
-  public User registerAsMember(Long userId, String fullName) {
+  public User registerAsMember(Long userId) {
     User user =
         userRepository
             .findById(userId)
--- a/src/main/java/ee/tuleva/onboarding/user/UserService.java
+++ b/src/main/java/ee/tuleva/onboarding/user/UserService.java
@@ -1,5 +1,6 @@
 package ee.tuleva.onboarding.user;
 
+import ee.tuleva.onboarding.user.exception.DuplicateEmailException;
 import ee.tuleva.onboarding.user.exception.UserAlreadyAMemberException;
 import ee.tuleva.onboarding.user.member.Member;
 import ee.tuleva.onboarding.user.member.MemberRepository;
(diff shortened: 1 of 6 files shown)


src/main/java/ee/tuleva/onboarding/user/UserService.java
```diff
@@ -82,4 +87,9 @@ public User save(User user) {
     log.info("Saving user: userId={}", user.getId());
     return userRepository.save(user);
   }
+
+  public boolean isEmailExist(String personalCode, String email) {
+    Optional<User> existingUser = userRepository.findByEmail(email);
+    return existingUser.isPresent() && !personalCode.equals(existingUser.get().getPersonalCode());
+  }
```
ErkoRisthein, line 94: Why is this a public method? I.e. does it necessarily need to be as part of the public interface of this class or can it be private? Is it used anywhere outside of this class? Should it ever be used outside of the `UserService`?

ErkoRisthein: Also, the name of this method is not grammatically correct in English. I usually like to Google phrases to see how many responses I get: https://www.google.com/search?q=%22is+email+exist%22 shows 10 responses.
I would name it something like "isExistingEmail" or "isDuplicateEmail"

ErkoRisthein:
What's the current behavior and the new behavior?

src/main/java/ee/tuleva/onboarding/user/UserService.java
ErkoRisthein: Oh, it's used in the `UserDetailsUpdater`, got it.

MatsJohanson: Yeah, the mandatory Java **is** prefix always creates interesting situations, since we can't use **does**, so in my past lives I've just ignored the grammar. I'll change it to something more sensible.

MatsJohanson:
> What's the current behavior and the new behavior?

I should have been more verbose in the PR description.

Prerequisites: Pension centre has several users with the same e-mail, one of which is already our user. The second user tries to log in.

Previously: 
1. we fetch the pension centre data, try to update the user table and get an [sql exception](https://sentry.io/organizations/tuleva/issues/2053127545/?project=146944)
2. user sees "Oi, sisselogimisel tekkis viga. Palun proovi uuesti."
3. they are stuck and unable to continue

Now: 
1. we fetch the pension centre data
2. check if the e-mail exists among our users
3. leave e-mail empty if so
4. direct the user to contact details where they get a proper error message ("E-mail on juba kasutuses") when trying to enter the same e-mail
5. they can enter another e-mail and use Tuleva

PR #468, opened 2021-11-09
Fix mobile id mandate signing for people with slow connections

MatsJohanson:
When simulating a slow 3G connection in Chrome, it's possible to reproduce the "Mobile session not found" error. This means that we have a race condition between async front-end requests and Spring session security. Each request gets a new session ID and if there are several requests before the previous one returns, our custom session storage becomes inaccessible.

This PR removes the dependency on MobileIdSession, reducing the chances that it gets invalidated before the user reaches the mandate signing step.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/auth/idcard/IdCardSession.java
+++ b/src/main/java/ee/tuleva/onboarding/auth/idcard/IdCardSession.java
@@ -1,6 +1,7 @@
 package ee.tuleva.onboarding.auth.idcard;
 
 import ee.tuleva.onboarding.auth.principal.Person;
+import java.io.Serial;
 import java.io.Serializable;
 import lombok.Builder;
 import lombok.Data;
@@ -11,10 +12,15 @@ import lombok.RequiredArgsConstructor;
 @Builder
 public class IdCardSession implements Person, Serializable {
 
-  private static final long serialVersionUID = -111852724795571891L;
+  @Serial private static final long serialVersionUID = -111852724795571891L;
 
   public final String firstName;
   public final String lastName;
   public final String personalCode;
   public final IdDocumentType documentType;
+
+  @Override
+  public String getPhoneNumber() {
+    return null;
+  }
 }
--- a/src/main/java/ee/tuleva/onboarding/auth/ocsp/CheckCertificateResponse.java
+++ b/src/main/java/ee/tuleva/onboarding/auth/ocsp/CheckCertificateResponse.java
@@ -1,6 +1,7 @@
 package ee.tuleva.onboarding.auth.ocsp;
 
 import ee.tuleva.onboarding.auth.principal.Person;
+import java.io.Serial;
 import java.io.Serializable;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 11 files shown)


src/main/java/ee/tuleva/onboarding/auth/principal/Person.java
```diff
@@ -7,4 +7,6 @@
   String getFirstName();
 
   String getLastName();
+
+  String getPhoneNumber();
```
ErkoRisthein, line 11: I would not add the `phoneNumber` field to the `Person` since it's totally valid that a person might not have a phone number. You can see the problem with this change also in the places where you're forced to add this Override:
```
  @Override
  public String getPhoneNumber() {
    return null;
  }
  ```
This is a clear indication that we have a problem with our OOP design.

I would maybe introduce a new interface `PersonWithPhone extends Person` or create a separate interface `WithPhoneNumber`, so you could do stuff like `AuthenticatedPerson implements Person, WithPhoneNumber`.

Any other thoughts on how to model this properly in OOP?

File: src/main/java/ee/tuleva/onboarding/auth/idcard/IdCardSession.java [hunk 2]
```diff
@@ -11,10 +12,15 @@
...
 
   public final String firstName;
   public final String lastName;
   public final String personalCode;
   public final IdDocumentType documentType;
+
+  @Override
+  public String getPhoneNumber() {
+    return null;
+  }
```
ErkoRisthein, line 25: We should not have the phoneNumber as part of the top-level interface if it's not relevant for _all_ the implementation classes.

src/main/java/ee/tuleva/onboarding/auth/ocsp/CheckCertificateResponse.java
```diff
@@ -11,9 +12,14 @@
...
+  @Serial private static final long serialVersionUID = 333351175769353073L;
 
   private final String firstName;
   private final String lastName;
   private final String personalCode;
+
+  @Override
+  public String getPhoneNumber() {
+    return null;
+  }
```
ErkoRisthein, line 24: We should not have the phoneNumber as part of the top-level interface if it's not relevant for all the implementation classes.

src/main/java/ee/tuleva/onboarding/auth/smartid/SmartIdSession.java
```diff
@@ -24,4 +25,9 @@ public void setAuthenticationIdentity(AuthenticationIdentity identity) {
     lastName = identity.getSurname();
     country = identity.getCountry();
   }
+
+  @Override
+  public String getPhoneNumber() {
+    return null;
+  }
```
ErkoRisthein, line 32: We should not have the phoneNumber as part of the top-level interface if it's not relevant for all the implementation classes.

ErkoRisthein:
Good job debugging & fixing this issue! We need to figure out the correct OOP design here and we're good to go.

src/main/java/ee/tuleva/onboarding/auth/principal/Person.java
ErkoRisthein: Maybe using the `Strategy` pattern could work in this case? https://medium.com/@shantanutripathi/strategy-pattern-the-first-design-pattern-to-understand-634e9adab243

ErkoRisthein:
I think this might be a problem with Smart-ID as well?

MatsJohanson:
> I think this might be a problem with Smart-ID as well?

Smart-ID doesn't pull out the SmartID session from the session store before starting signing so there is no chance of the async chain messing up its state between checks.

MatsJohanson:
Thanks for the refacto branch, I was already offline for the day last night. I'll merge this branch as all the comments are handled in yours.

PR #469, opened 2021-11-10
Refactor Person interface

ErkoRisthein:


src/main/java/ee/tuleva/onboarding/auth/principal/PrincipalService.java
```diff
@@ -17,7 +18,7 @@
 
   private final UserService userService;
 
-  public AuthenticatedPerson getFrom(Person person) {
+  public AuthenticatedPerson getFrom(Person person, Supplier<Optional<String>> phoneSupplier) {
```
MatsJohanson, line 21: Why a supplier instead of just an `Optional<String> phoneNumber`?

MatsJohanson (commented):
Looks cleaner than just adding `return null` methods, thanks :bow: . We can simplify it further by skipping the supplier function.

src/main/java/ee/tuleva/onboarding/auth/principal/PrincipalService.java
ErkoRisthein: Good catch, feel free to simplify it :)

MatsJohanson (approved):

PR #474, opened 2021-11-22
Take null e-mails into account when creating a user

MatsJohanson:
Our current code didn't take into account that pulled e-mails can be "null" from Pensionikeskus for some users. Now we don't try to compare "null" e-mails to find out if they are a duplicate of any existing e-mails.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/user/UserService.java
+++ b/src/main/java/ee/tuleva/onboarding/user/UserService.java
@@ -35,7 +35,7 @@ public class UserService {
     return userRepository.save(user);
   }
 
-  public User updateUser(String personalCode, String email, String phoneNumber) {
+  public User updateUser(String personalCode, Optional<String> email, String phoneNumber) {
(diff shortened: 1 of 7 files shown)


src/main/java/ee/tuleva/onboarding/user/UserService.java
```diff
@@ -44,7 +44,7 @@ public User updateUser(String personalCode, String email, String phoneNumber) {
         findByPersonalCode(personalCode)
             .map(
                 existingUser -> {
-                  existingUser.setEmail(email);
+                  existingUser.setEmail(email.orElse(null));
```
ErkoRisthein, line 47: Any risk of overwriting existing emails in the database with nulls?

MatsJohanson: Nein. We're calling this from two places: 
1. front-end where the e-mail presence is validated
2. `updateContactDetails` where we're only running the code if both e-mail and phone number are empty so nothing to lose

PR #494, opened 2022-01-03
Fix early withdrawal fulfillment date

MatsJohanson:
Expectation: When an application is submitted in October 2021, its fulfillment date should be in May 2022. Currently we are showing it to be September 2022. 

This PR rectifies the given situation - we're now taking the application year into account when calculating its deadlines.


ErkoRisthein:
Any unit test that would replicate this bug?

Thanks for fixing my bugs 🙇

MatsJohanson:
> Any unit test that would replicate this bug?
> 
> Thanks for fixing my bugs bow

Good call. I added the test-case here: https://github.com/TulevaEE/onboarding-service/compare/b5602414b5a7...6f36101527ac

ErkoRisthein:
🙏

PR #496, opened 2022-01-04
Replace Springfox with Springdoc and update spring-cloud-sleuth

MatsJohanson:
Springfox has not had an update for over a year and they have conflicts with Spring Boot upgrades. The prevalent notion is to migrate to Springdoc, which is a more maintained Swagger UI provider.

With the addition of `springdoc-openapi-security`, we can also get rid of all @ApiIgnore usages - AuthenticatedPerson parameters will now be ignored automatically.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
+++ b/src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
@@ -1,7 +1,7 @@
 package ee.tuleva.onboarding.comparisons;
 
 import ee.tuleva.onboarding.auth.principal.AuthenticatedPerson;
-import io.swagger.annotations.ApiOperation;
+import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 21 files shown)


src/main/java/ee/tuleva/onboarding/comparisons/FundComparisonController.java
```diff
@@ -25,10 +24,18 @@
 
   private final FundComparisonCalculatorService fundComparisonCalculatorService;
 
-  @ApiOperation(value = "Compare the current user's return to estonian and world average")
+  private static Instant parseInstant(String format) {
```
ErkoRisthein, line 27: According to Clean Code, the order of methods in a class should be so that high-level details come at the top of the class and as you scroll down you get to more lower-level details. And methods are ordered so they are close to the ones that call them. The caller first, callee second. You can read more about this philosophy from this summary: https://softwareengineering.stackexchange.com/a/199317

PR #541, opened 2022-04-27
Remove fund comparison

ErkoRisthein:
Remove some leftover deprecated code. Use Returns instead of the old FundComparison.


ErkoRisthein:
@MatsJohanson Can we enable codecov again?

PR #543, opened 2022-04-29
Re-enable code coverage measuring

MatsJohanson:
Re-add codecov using their official CircleCI orb, which is no longer arbitrarily downloading a bash file and running it.


ErkoRisthein (approved):
❤️

ErkoRisthein:
It still downloads the bash file & executes it though, if you look at the orb source code: https://circleci.com/developer/orbs/orb/codecov/codecov
It also validates the PGP signature, so it's better than nothing, but it's still executing _arbitrary_ code. 😞

PR #577, opened 2022-06-28
https://app.asana.com/0/229163390356948/1202508354498827

gnortsmralien:
Remove duplicate authentication from access token table and add unique constraint


src/main/resources/db/migration/V1_54__restrict_duplicate_authentication_id.sql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM oauth_access_token WHERE authentication_id IN
+                                     (SELECT authentication_id
+                                      FROM oauth_access_token
+                                      GROUP BY authentication_id
+                                      HAVING count(*) > 1);
+
+ALTER TABLE oauth_access_token ADD CONSTRAINT unique_authentication_id UNIQUE (authentication_id);
```
ErkoRisthein, line 7: The unique constraint naming convention for postgres would be `{tableName}_{columnName}_{suffix}` 

i.e. `oauth_access_token_authentication_id_key` in this case.

See https://stackoverflow.com/a/4108266

ErkoRisthein (approved):

PR #597, opened 2022-08-15
Refactor audit_log into event_log and create TrackableEvent domain

ErkoRisthein:


src/main/resources/db/migration/V1_57__update_audit_log_data_field_type.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE audit_log
```
v3rm0n, line 1: Downtime during release

ErkoRisthein: How can we know how long?

v3rm0n: During the deployment from the time the new node starts up the application until the old node is drained of all connections.

PR #624, opened 2022-09-26
Adds a payment status

JordanValdma:


src/main/java/ee/tuleva/onboarding/payment/PaymentProviderCallbackService.java
```diff
@@ -43,6 +43,7 @@ public void processToken(String serializedToken) {
               .amount(amount)
               .internalReference(internalReference.getUuid())
               .user(user)
+              .status(PaymentStatus.PENDING)
```
ErkoRisthein, line 46: There might be a different status in the JWT token coming from Montonio

ErkoRisthein: We should probably double check that the Montonio payment status is `finalized`

JordanValdma: Done, thanks

src/main/java/ee/tuleva/onboarding/payment/PaymentProviderCallbackService.java
```diff
@@ -35,20 +35,29 @@ public void processToken(String serializedToken) {
...
               .build();
 
       paymentRepository.save(paymentToBeSaved);
     }
   }
 
+  private boolean paymentWithReferenceExists(PaymentReference internalReference) {
+    return paymentRepository.findByInternalReference(internalReference.getUuid()).isEmpty();
+  }
```
ErkoRisthein, line 55: `paymentWithReferenceExists == findByInternalReference isEmpty()`? Isn't it contradictory?

If it's empty then it _doesn't_ exist.

JordanValdma: fixed, thanks

ErkoRisthein (approved):

PR #625, opened 2022-09-26
Use accessKey to get

JordanValdma:


src/main/java/ee/tuleva/onboarding/payment/PaymentProviderConfiguration.java
```diff
@@ -16,19 +16,20 @@
...
   }
 
-  public PaymentProviderBank getPaymentProviderBank(String bic) {
-    return banksByBic.get(bic);
+  public PaymentProviderBank getPaymentProviderBank(String accessKey) {
+    return banksByAccessKey.get(accessKey);
   }
 
   @PostConstruct
   private void mapByBic() {
```
ErkoRisthein, line 30: `mapByAccessKey` would be a proper name

JordanValdma: I disagree, map is a type. No point in referring a type in the variable

JordanValdma: Oh, I get it

ErkoRisthein (approved):

PR #674, opened 2022-11-11
Remove session from SmartID authentication

v3rm0n:


src/main/java/ee/tuleva/onboarding/auth/session/GenericSessionStore.java
```diff
@@ -1,33 +1,49 @@
...
+import javax.servlet.http.HttpSession;
 import org.springframework.stereotype.Component;
-import org.springframework.web.context.annotation.SessionScope;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
 
 @Component
-@SessionScope
-public class GenericSessionStore implements Serializable {
+public class GenericSessionStore {
```
ErkoRisthein, line 13: What changed here? Not a session scoped bean anymore?

src/main/resources/application.yml
```diff
@@ -40,6 +40,7 @@ spring:
     store-type: none
     jdbc:
       initialize-schema: never
+  cloud.mvc.token-relay.enabled: false
```
ErkoRisthein, line 43: What does it affect?

src/main/java/ee/tuleva/onboarding/auth/session/GenericSessionStore.java
v3rm0n: Yes, injecting a session scoped bean will create a session automatically which we do not want

src/main/resources/application.yml
v3rm0n: It uses a session so it forces one to be created. It didn't seem to be used anywhere, at least downstream OAuth2 calls still work fine even when this is disabled.

ErkoRisthein (approved):

PR #675, opened 2022-11-11
Roi cash calculation

JordanValdma:


src/main/java/ee/tuleva/onboarding/comparisons/returns/provider/FundReturnProvider.java
```diff
@@ -35,10 +35,15 @@ public Returns getReturns(Person person, Instant startTime, Integer pillar) {
             .map(
                 key ->
                     new SimpleEntry<>(
-                        key, rateOfReturnCalculator.getRateOfReturn(accountOverview, key)))
+                        key, rateOfReturnCalculator.getReturnRateAndAmount(accountOverview, key)))
```
ErkoRisthein, line 38: Can we just call it `Return`? i.e. `getReturn(...)`

src/main/java/ee/tuleva/onboarding/comparisons/returns/Returns.java
```diff
@@ -19,7 +20,9 @@ public class Returns {
   public static class Return {
 
     String key;
-    double value; // TODO: migrate to BigDecimal
+    BigDecimal rate;
+
+    BigDecimal amount;
```
ErkoRisthein, line 25: Currency as well?

src/main/java/ee/tuleva/onboarding/comparisons/returns/provider/FundReturnProvider.java
JordanValdma: Thanks, fixed

src/main/java/ee/tuleva/onboarding/comparisons/returns/Returns.java
JordanValdma: Thanks! Added

src/main/java/ee/tuleva/onboarding/comparisons/returns/ReturnRateAndAmount.java
```diff
@@ -0,0 +1,6 @@
+package ee.tuleva.onboarding.comparisons.returns;
+
+import ee.tuleva.onboarding.currency.Currency;
+import java.math.BigDecimal;
+
+public record ReturnRateAndAmount(BigDecimal rate, BigDecimal amount, Currency currency) {}
```
ErkoRisthein, line 6: I usually prefer to avoid using the word "and" in any method names and class names. It's usually a code smell (a violation of the Single Responsibility Principle). I.e. a class that does more than one thing (and could be split into 2) or a method that does more than 1 thing (and could be split).

ErkoRisthein (approved):

src/main/java/ee/tuleva/onboarding/comparisons/returns/ReturnRateAndAmount.java
JordanValdma: good point

JordanValdma: Fixed

PR #841, opened 2024-03-04
Refreshable tokens

JordanValdma:


src/main/java/ee/tuleva/onboarding/auth/AccessAndRefreshToken.java
```diff
@@ -0,0 +1,7 @@
+package ee.tuleva.onboarding.auth;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record AccessAndRefreshToken(
```
ErkoRisthein, line 5: "And" in a class name (or method name) is a code smell - it hints at a Single Responsibility violation ("and" means, by definition, that the class does two things, not one). I would rename this to just "Token", "AuthToken", "TokenPair" or something simple like that.

src/main/java/ee/tuleva/onboarding/auth/AuthService.java
```diff
@@ -32,12 +38,37 @@ public String authenticate(GrantType grantType, String authenticationHash) {
       eventPublisher.publishEvent(
           new BeforeTokenGrantedEvent(this, authenticatedPerson, grantType));
 
-      String jwtToken = jwtTokenUtil.generateToken(authenticatedPerson, authorities);
+      String jwtToken = jwtTokenUtil.generateAccessToken(authenticatedPerson, authorities);
```
ErkoRisthein, line 41: This should be named `accessToken`

src/main/java/ee/tuleva/onboarding/auth/ExpiredRefreshJwtException.java
```diff
@@ -0,0 +1,8 @@
+package ee.tuleva.onboarding.auth;
+
+public class ExpiredRefreshJwtException extends RuntimeException {
```
ErkoRisthein, line 3: `ExpiredRefreshTokenException` would read better and would be implementation-agnostic (i.e. the token could be in JWT format, or some other format as well)

src/main/java/ee/tuleva/onboarding/auth/jwt/JwtAuthorizationFilter.java
```diff
@@ -36,30 +39,42 @@ protected void doFilterInternal(
       }
       try {
         if (SecurityContextHolder.getContext().getAuthentication() == null) {
-          AuthenticatedPerson principal =
-              principalService.getFrom(
-                  jwtTokenUtil.getPersonFromToken(jwtToken),
-                  jwtTokenUtil.getAttributesFromToken(jwtToken));
+          TokenType tokenType = jwtTokenUtil.getTypeFromToken(jwtToken);
+          if (tokenType == TokenType.ACCESS) {
```
ErkoRisthein, line 43: I would create a validation block at the beginning here instead of indenting the whole business logic by 1 level.
For example:
```
if (tokenType != TokenType.ACCESS) {
  return; // do nothing
}
// business logic
```
instead of:
```
if (tokenType == TokenType.ACCESS) {
  // business logic
}
// do nothing
```

src/main/java/ee/tuleva/onboarding/auth/jwt/JwtTokenUtil.java
```diff
@@ -60,14 +59,18 @@ public Person getPersonFromToken(String token) {
   }
 
   private Claims getAllClaimsFromToken(String token) {
+    // When token is expired, this throws ExpiredJwtException, which is handled in
```
ErkoRisthein, line 62: I'd prefer `ExpiredTokenException` - that wouldn't leak implementation details (that we're currently using a token in JWT format).

src/main/java/ee/tuleva/onboarding/error/ErrorHandlingControllerAdvice.java
```diff
@@ -62,4 +66,20 @@ public ResponseEntity<Object> handleErrors(ErrorsResponseException exception) {
     log.error("{}", exception.toString());
     return new ResponseEntity<>(exception.getErrorsResponse(), BAD_REQUEST);
   }
+
+  @ExceptionHandler(ExpiredJwtException.class)
+  public ResponseEntity<Object> handleErrors(ExpiredJwtException exception) {
+    log.debug("ExpiredJwtException {}", exception.toString());
+    return new ResponseEntity<>(
+        JwtTokenUtil.getExpiredTokenErrorResponse(), HttpStatus.UNAUTHORIZED);
```
ErkoRisthein, line 74: I would hardcode the error code and message in this class, not in the `JwtTokenUtil` class

src/main/java/ee/tuleva/onboarding/config/SecurityConfiguration.java
```diff
@@ -56,10 +57,15 @@ public SecurityFilterChain securityFilterChain(
...
+            "/authenticate",
+            "/oauth/token",
+            "/oauth/refresh-token",
+            "/idLogin",
+            "/notifications/payments",
+            "/v1/**")
         .and()
         .sessionManagement()
-        .sessionCreationPolicy(SessionCreationPolicy.NEVER)
+        .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
```
ErkoRisthein, line 68: What's the difference between NEVER and STATELESS?

src/main/java/ee/tuleva/onboarding/config/SecurityConfiguration.java
```diff
@@ -33,6 +33,7 @@ public SecurityFilterChain securityFilterChain(
             "/v3/api-docs/**",
             "/authenticate",
             "/oauth/token",
+            "/oauth/refresh-token",
```
ErkoRisthein, line 36: Let's try to get rid of all `/oauth` endpoints - we're not using oauth anymore. Let's make the endpoints generic and implementation agnostic (i.e. just `/v1/tokens` endpoint, for example)

src/main/java/ee/tuleva/onboarding/auth/session/GenericSessionStore.java
```diff
@@ -1,49 +1,57 @@
...
+    return repository
+        .findByUserIdAndAttributeName(userId, clazz.getName())
+        .flatMap(
+            attribute -> {
+              try (ObjectInput in =
+                  new ObjectInputStream(new ByteArrayInputStream(attribute.getAttributeValue()))) {
+                @SuppressWarnings("unchecked")
+                T result = (T) in.readObject();
+                return Optional.of(result);
+              } catch (IOException | ClassNotFoundException e) {
```
ErkoRisthein, line 52: We should never catch `ClassNotFoundException`!! This is a classpath configuration error that should be resolved

src/main/java/ee/tuleva/onboarding/auth/session/GenericSessionStore.java
```diff
@@ -1,49 +1,57 @@
...
-import org.springframework.web.context.request.ServletRequestAttributes;
 
 @Component
+@RequiredArgsConstructor
 public class GenericSessionStore {
 
-  private static final String GENERIC_SESSION_STORE_ATTRIBUTES =
-      GenericSessionStore.class.getName() + ".attributes";
+  private final SessionAttributeRepository repository;
+  private final AuthenticationHolder authenticationHolder;
```
ErkoRisthein, line 14: This is not a "generic session store" anymore. It depends on the User domain, database, SessionAttributeRepository. I would rename it to just "SessionStore" or something like that

src/main/java/ee/tuleva/onboarding/auth/AuthService.java
```diff
@@ -32,12 +38,37 @@ public String authenticate(GrantType grantType, String authenticationHash) {
...
     }
     return null;
   }
+
+  public AccessAndRefreshToken refreshToken(String refreshToken) {
+
+    try {
+      TokenType tokenType = jwtTokenUtil.getTypeFromToken(refreshToken);
+      if (tokenType != TokenType.REFRESH) {
+        throw new BadRequestException("Only refresh token is allowed for refresh request.");
```
ErkoRisthein, line 57: Why are we using `javax.ws.rs.BadRequestException`? Let's use standard Java API `IllegalArgumentException`

src/test/groovy/ee/tuleva/onboarding/auth/jwt/JwtAuthorizationFilterSpec.groovy
```diff
@@ -119,4 +127,27 @@ class JwtAuthorizationFilterSpec extends Specification {
     1 * filterChain.doFilter(request, response)
   }
 
+  def "does not accept refresh tokens"() {
```
ErkoRisthein, line 130: Should also test that it doesn't accept other types of tokens (i.e. "asdfg") - to get more confidence that we _only_ accept access_tokens. For example, right now it would not catch the bug in case we would accidentally assert:

```
if (tokenType == TokenType.REFRESH) {
  return;
}
// accept all other token types
```

src/test/groovy/ee/tuleva/onboarding/auth/AuthControllerSpec.groovy
```diff
@@ -108,6 +110,56 @@ class AuthControllerSpec extends BaseControllerSpec {
...
+  def "Refresh Access Token: handle expired refresh token"() {
+    given:
+        String expiredRefreshToken = "expiredRefreshToken"
+        authService.refreshToken(expiredRefreshToken) >> { throw new ExpiredRefreshJwtException() }
+
+    when:
+        MockHttpServletResponse response = mockMvc
+            .perform(post("/oauth/refresh-token")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content("{\"refresh_token\":\"${expiredRefreshToken}\"}")).andReturn().response
```
ErkoRisthein, line 159: You can use triple quotes in groovy instead of escaping:  `"""{"refresh_token":"${expiredRefreshToken}"}"""`

src/main/java/ee/tuleva/onboarding/auth/jwt/JwtAuthorizationFilter.java
```diff
@@ -36,30 +39,42 @@ protected void doFilterInternal(
...
+                new WebAuthenticationDetailsSource().buildDetails(request));
 
-          SecurityContextHolder.getContext().setAuthentication(authenticationToken);
+            SecurityContextHolder.getContext().setAuthentication(authenticationToken);
+          }
         }
       } catch (ExpiredJwtException e) {
         logger.info("JWT Token is expired");
+        respondWithTokenExpired(response);
+        return;
```
ErkoRisthein, line 66: Is it OK to return in a Security Filter? shouldn't it continue with the whole filter chain?

ErkoRisthein: Ok, from the [docs](https://docs.oracle.com/javaee/6/api/javax/servlet/Filter.html#doFilter%28javax.servlet.ServletRequest,%20javax.servlet.ServletResponse,%20javax.servlet.FilterChain%29) I see that:
- **Either** invoke the next entity in the chain using the FilterChain object (chain.doFilter()),
- **or** not pass on the request/response pair to the next entity in the filter chain to block the request processing

src/test/groovy/ee/tuleva/onboarding/error/ErrorHandlingControllerAdviceSpecification.groovy
```diff
@@ -0,0 +1,104 @@
...
+import spock.lang.Specification
+import org.springframework.http.HttpStatus
+import ee.tuleva.onboarding.error.response.ErrorResponseEntityFactory
+import ee.tuleva.onboarding.error.response.ErrorsResponse
+import org.springframework.http.ResponseEntity
+import ee.tuleva.onboarding.mandate.exception.IdSessionException
+import io.jsonwebtoken.ExpiredJwtException
+import ee.tuleva.onboarding.auth.ExpiredRefreshJwtException
+
+class ErrorHandlingControllerAdviceSpecification extends Specification {
```
ErkoRisthein, line 15: Should be called just "...Spec"

src/main/java/ee/tuleva/onboarding/auth/jwt/JwtTokenUtil.java
```diff
@@ -78,7 +81,29 @@ public String generateToken(
...
+        .setExpiration(Date.from(clock.instant().plus(ACCESS_TOKEN_VALIDITY)))
+        .signWith(signingKey)
+        .compact();
+  }
+
+  public String generateRefreshToken(
+      AuthenticatedPerson person, Collection<? extends GrantedAuthority> authorities) {
+    return Jwts.builder()
+        .setClaims(
+            Map.of(
```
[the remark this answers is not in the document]
JordanValdma: Explored that route also. Yes, refreshing the token without that data is possible, but for simplicity it is in the token.  You would need to "log" the user in again to get that data, which is more complex

src/main/java/ee/tuleva/onboarding/error/ErrorHandlingControllerAdvice.java
JordanValdma: The error is used in two different places, so for a clear definition, I extracted it

src/main/java/ee/tuleva/onboarding/config/SecurityConfiguration.java
JordanValdma: Feature | NEVER | STATELESS
-- | -- | --
Creates HttpSession | No | No
Uses Existing HttpSession | Yes | No
Statefulness | Minimal (uses existing sessions) | Stateless
Ideal Use Case | Applications that mix stateful and stateless operations | Purely stateless applications like REST APIs

src/main/java/ee/tuleva/onboarding/auth/session/GenericSessionStore.java
JordanValdma: I think the author intended "generic" as a " generic type"

src/test/groovy/ee/tuleva/onboarding/auth/jwt/JwtAuthorizationFilterSpec.groovy
JordanValdma: TokenType is an enum, that field is typed and can only have values ACCESS, REFRESH

src/test/groovy/ee/tuleva/onboarding/error/ErrorHandlingControllerAdviceSpecification.groovy
JordanValdma: There is also ErrorHandlingControllerSpec, with a different test setup. I did not want to delete it

src/main/java/ee/tuleva/onboarding/auth/ExpiredRefreshJwtException.java
v3rm0n: FYI https://projectlombok.org/features/experimental/StandardException

src/main/resources/db/migration/V1_78__session_attribute.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS session_attribute (
+  id              BIGSERIAL PRIMARY KEY,
+  user_id         BIGINT REFERENCES users (id),
+  attribute_name  TEXT NOT NULL,
+  attribute_value BYTEA NOT NULL,
```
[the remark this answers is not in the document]
JordanValdma: JSON Serialisation issues with classes that are not under our control

PR #928, opened 2024-09-16
Fix deserializing MandateDetails when reading Mandate with Hibernate

snemvalts:
What this does:
* Removes custom deserializer registration, as they aren't needed, only `@JsonDeserialize` annotations matter.
* Adds another deserializer, this time for `MandateDetails` themselves, with some quite crazy logic:
  * Instantiates another `ObjectMapper`, disabling annotation inspector which would resolve to the deserializer itself causing an infinite loop.
  * Using that `ObjectMapper.treeToValue`, instantiates the `MandateDetails` itself 
* Adds `@SpringBootTest` integration test for validating that mandate can be read from database.


ErkoRisthein (approved):
Let's merge and fix it - we can talk about alternative solutions later (once it's working well in production)

PR #940, opened 2024-10-01
Add mandate batch signature logic

snemvalts:


src/main/java/ee/tuleva/onboarding/mandate/batch/MandateBatch.java
```diff
@@ -57,4 +58,8 @@ private void addFiles(CompositeMandateFileCreator fileCreator) {
     //    this.file = fileCreator.getContentFiles();
 
   }
+
+  public boolean isSigned() {
+    return this.status == SIGNED;
```
ErkoRisthein, line 63: No need for the `this` keyword: `return status == SIGNED;`

src/test/groovy/ee/tuleva/onboarding/mandate/MandateServiceSpec.groovy
```diff
@@ -201,7 +203,7 @@ class MandateServiceSpec extends Specification {
     def status = service.finalizeMobileIdSignature(sampleUser.id, sampleMandate.id, signatureSession, ENGLISH)
 
     then:
-    status == "OUTSTANDING_TRANSACTION"
+    status == OUTSTANDING_TRANSACTION
```
ErkoRisthein, line 206: Nice! Let's get rid of String constants and prefer enums instead

src/main/java/ee/tuleva/onboarding/mandate/batch/MandateBatch.java
snemvalts: Maybe there is a linter setting for it we could use? Similar to always prefer static on demand imports

PR #950, opened 2024-11-12
Add mandate batch auto-email logic, fix bug with second pillar suggestion not appearing in payment rate emails

snemvalts:


src/main/java/ee/tuleva/onboarding/mandate/MandateService.java
```diff
@@ -203,7 +203,7 @@ private MandateSignatureStatus handleUnsignedMandateIdCard(
 
   private void notifyAboutSignedMandate(User user, Mandate mandate, Locale locale) {
     applicationEventPublisher.publishEvent(
-        new AfterMandateSignedEvent(this, user, mandate, locale));
+        new AfterMandateSignedEvent(this, user, mandate, locale, false));
```
ErkoRisthein, line 206: What is this `false`? Let's avoid using boolean parameters, they are hard to understand

ErkoRisthein: Instead of true/false you could use `batchId` as a parameter and add a method `isPartOfBatch()`

src/main/java/ee/tuleva/onboarding/mandate/email/EmailVariablesAttachments.java
```diff
@@ -0,0 +1,76 @@
...
+
+    if (pillarSuggestion.isSuggestPaymentRate()) {
+      tags.add("suggest_payment_rate");
+    }
+    if (pillarSuggestion.isSuggestThirdPillar()) {
+      tags.add("suggest_3");
+    }
+    if (pillarSuggestion.isSuggestMembership()) {
+      tags.add("suggest_member");
+    }
```
ErkoRisthein, line 32: `suggest_2` tag missing

src/main/java/ee/tuleva/onboarding/mandate/email/MandateEmailSender.java
```diff
@@ -27,7 +31,24 @@ public void sendEmail(AfterMandateSignedEvent event) {
...
     PillarSuggestion pillarSuggestion =
         new PillarSuggestion(event.getUser(), contactDetails, conversion, paymentRates);
-    emailService.sendMandate(
-        event.getUser(), event.getMandate(), pillarSuggestion, event.getLocale());
+    if (!event.isPartOfBatch()) {
+      mandateEmailService.sendMandate(
+          event.getUser(), event.getMandate(), pillarSuggestion, event.getLocale());
+    } else {
+      log.info(
+          "Skipping mandate email for (id:{}) because it is part of batch",
```
ErkoRisthein, line 39: Prefer to use the following pattern for logging:
"Skipping mandate email because it is part of a batch: mandateId={}"

src/main/java/ee/tuleva/onboarding/mandate/MandateService.java
snemvalts: Yup, agreed. Moved it to mandate and added a method there

PR #971, opened 2024-12-10
Add withdrawal eligibility logic for III pillar with reduced tax 

snemvalts:
Adds withdrawal eligibility logic for III pillar with reduced tax (age is at least 55, III pillar first payment before 2021).

### DIFF
--- a/src/test/groovy/ee/tuleva/onboarding/mandate/batch/MandateBatchIntegrationTest.java
+++ b/src/test/groovy/ee/tuleva/onboarding/mandate/batch/MandateBatchIntegrationTest.java
@@ -115,8 +115,17 @@ public class MandateBatchIntegrationTest {
                     MandateDto.builder().details(aPartialWithdrawalMandateDetails).build()))
             .build();
 
+    var aWithdrawalEligibility =
+        WithdrawalEligibilityDto.builder()
+            .hasReachedEarlyRetirementAge(true)
+            .canWithdrawThirdPillarWithReducedTax(true)
+            .age(65)
+            .recommendedDurationYears(20)
+            .arrestsOrBankruptciesPresent(false)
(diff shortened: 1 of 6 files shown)


src/test/groovy/ee/tuleva/onboarding/mandate/batch/MandateBatchIntegrationTest.java
```diff
@@ -147,7 +147,7 @@ void testMandateCreationBeforeRetirementAge() throws Exception {
             .build();
 
     when(withdrawalEligibilityService.getWithdrawalEligibility(any()))
-        .thenReturn(new WithdrawalEligibilityDto(false, 30, 55, false));
+        .thenReturn(new WithdrawalEligibilityDto(false, false, 30, 55, false));
```
ErkoRisthein, line 150: Any way to avoid boolean fields like this? `new WithdrawalEligibilityDto(false, false, 30, 55, false)` is not really understandable what they represent

snemvalts: It's the unfortunate combination of record (no builders) + non-fleshed out abstraction. 
As we get more information on how other withdrawals are going to look like, we can refactor this DTO to something better. 
But even then we would probably write dead code we won't use to fill the "empty space" in the abstraction to make it complete.

For intellij, these boolean flags have the little name decorations which make it more readable Screenshot 2024-12-10 at 11 31 23

snemvalts: Ah, nevermind. Builders can be used with record. But then problems with potentially nullable fields start, forgetting them etc.

PR #973, opened 2024-12-12
Add enum for epis cache names

snemvalts:
Epis cache names to enum


src/main/java/ee/tuleva/onboarding/epis/EpisService.java
```diff
@@ -253,4 +269,28 @@ private String userJwtToken() {
   private String serviceJwtToken() {
     return jwtTokenUtil.generateServiceToken();
   }
+
+  @FieldNameConstants(onlyExplicitlyIncluded = true)
```
snemvalts, line 273: https://stackoverflow.com/questions/77598934/can-i-use-enum-in-cacheable

PR #1044, opened 2025-02-25
Check last contact details update time from EPIS, add relevant AML contact details check in that case

snemvalts:


src/main/java/ee/tuleva/onboarding/aml/AmlCheckService.java
```diff
@@ -28,20 +31,35 @@ public void addCheckIfMissing(Person person, AmlCheckAddCommand command) {
   }
 
   public List<AmlCheckType> getMissingChecks(Person person) {
-    final var checks =
-        stream(AmlCheckType.values()).filter(AmlCheckType::isManual).collect(toList());
-    final var existingChecks =
+    final var doneChecksTypes =
```
snemvalts, line 34: Rename

snemvalts:
Not actually needed, root cause is contact details update failing in EPIS for some reason while being marked as complete

snemvalts:
Will try a different approach

PR #1155, opened 2025-05-06
Hotfix auto-emails after historical data import

ErkoRisthein:


JordanValdma (approved):

src/main/java/ee/tuleva/onboarding/notification/email/auto/AutoEmailSender.java
```diff
@@ -42,6 +42,16 @@ public void sendMonthlyEmails() {
           startDate,
           endDate);
       final var emailablePeople = autoEmailRepository.fetch(startDate, endDate);
+
+      // Sanity check to make sure we don't accidentally send too many emails
+      if (emailablePeople.size() > 100) {
```
snemvalts, line 47: this will start throwing, as we have exceed this in many cases, need to filter by people we won't send emails to because of 4 month threshold

PR #1166, opened 2025-05-15
Add swedbank gateway XSD and generate classes based on them

snemvalts:


build.gradle.kts
```diff
@@ -99,6 +101,8 @@ dependencies {
     implementation("org.eclipse.persistence:org.eclipse.persistence.moxy:4.0.6")
     implementation("jakarta.xml.bind:jakarta.xml.bind-api:4.0.2")
 
+    xjc("org.glassfish.jaxb:jaxb-xjc:4.0.4")
```
ErkoRisthein, line 104: Why not use the latest version?

PR #1180, opened 2025-06-03
Add /listings API, fix DI problems with @WebMvcTest, fix tests

ErkoRisthein:


src/main/java/ee/tuleva/onboarding/listing/ListingDto.java
```diff
@@ -0,0 +1,27 @@
+package ee.tuleva.onboarding.listing;
+
+import ee.tuleva.onboarding.currency.Currency;
+import jakarta.validation.constraints.*;
+import java.math.BigDecimal;
+import java.time.Instant;
+
+public record ListingDto(
+    @NotNull Long id,
```
snemvalts, line 9: If we go with serial, string would be better here, because IDs cannot be meaningfully be numbers (addition subtraction etc)

PR #1184, opened 2025-06-11
Only require III pillar AML checks for Tuleva III pillar customer's mandate

snemvalts:


src/main/java/ee/tuleva/onboarding/aml/AmlAutoChecker.java
```diff
@@ -67,14 +70,30 @@ public void beforeMandateCreated(BeforeMandateCreatedEvent event) {
...
 
     if (!amlService.allChecksPassed(user, pillar)) {
       throw AmlChecksMissingException.newInstance();
     }
   }
 
+  private Boolean isThirdPillarMandateAmlCheckRequired(User user, Mandate mandate) {
+    var conversion = userConversionService.getConversion(user).getThirdPillar();
+    var isTulevaThirdPillarClient =
+        conversion.isPartiallyConverted() || conversion.isFullyConverted();
```
ErkoRisthein, line 86: `conversion.isPartiallyConverted()` would suffice, but not harm doing both checks either.

src/test/groovy/ee/tuleva/onboarding/aml/AmlAutoCheckerTest.java
```diff
@@ -214,18 +224,102 @@ void beforeMandateCreated_noSanctionAndPepCheck_forNonThirdPillar_whenChecksPass
...
+    BeforeMandateCreatedEvent mockEvent = mock(BeforeMandateCreatedEvent.class);
+    var mandate = samplePartialWithdrawalMandate(aThirdPillarPartialWithdrawalMandateDetails);
+    Integer pillar = mandate.getPillar();
+    assertEquals(3, pillar);
+
+    when(mockEvent.getUser()).thenReturn(mockUser);
+    when(mockEvent.getMandate()).thenReturn(mandate);
+    when(mockEvent.getPillar()).thenReturn(pillar);
+    when(mockEvent.isThirdPillar()).thenReturn(true);
+    when(amlService.allChecksPassed(mockUser, pillar)).thenReturn(true);
```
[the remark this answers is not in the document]
snemvalts: As it is an unit test of `AmlAutoChecker` and it's asking it from external class `AmlService` it should be OK to mock. 
Alternatively `allChecksPassed` could be a static method, then this mock wouldn't be needed, but the method has a DI injected event publisher dependency at the moment.

src/test/groovy/ee/tuleva/onboarding/aml/AmlServiceTest.java
```diff
@@ -424,6 +435,60 @@ void allChecksPassed_evaluatesThirdPillar(
...
+            Arguments.of(notConverted(), thirdPillarMandate(), true),
+            Arguments.of(notConverted(), sampleMandate(), false),
+            Arguments.of(
+                notConverted(),
+                samplePartialWithdrawalMandate(aThirdPillarPartialWithdrawalMandateDetails),
+                false),
+            Arguments.of(
+                notConverted(),
+                samplePartialWithdrawalMandate(aPartialWithdrawalMandateDetails),
+                false));
```
ErkoRisthein, line 474: Can we use an enum instead of true/false here for better readability?

PR #1273, opened 2025-09-25
Capital executor logic for clamping to total when difference is less than 1 cent

snemvalts:


JordanValdma (approved):
All rounding should be down

PR #1373, opened 2025-12-03
Do not extend ApplicationEvent in every Event

vootelerotov:
This was needed at one point, but no longer is.

Allows for events to be records.


vootelerotov:
Boy-scouting, let's deal with this after sprint.

PR #1565, opened 2026-04-20
Cap TKF NAV history to safe publication date (T+1)

stenehrlich-tuleva:
## Summary
TKF NAV appears in the database at 15:20 but is officially published at 16:00 (and may be corrected before 16:00). This caps the public `/v1/funds/{isin}/nav` endpoint's end date to `fundNavProvider.safeMaxNavDate()` for the savings fund to prevent early publication.

Only affects TKF. Pension funds (TUK75, TUK00, TUV100) are not capped — their NAVs are already public when they hit our DB.

## Hardening from reviews
- **Use DB fund ISIN** for the savings-fund check and repository query, not the raw URL path — survives any future case-insensitive ISIN lookup.
- **`Cache-Control: no-store`** on both JSON and CSV responses so a CDN can't serve pre-16:00 cached bytes after publication.
- **Return empty list** when `startDate > endDate` after cap, instead of relying on repository semantics.

## Test plan
- [x] `FundServiceSpec` — 4 new tests (cap applies to savings, not to pension, empty on start>end, uses DB ISIN)
- [x] `FundControllerSpec` — asserts `Cache-Control: no-store` on both paths
- [ ] CircleCI passes
- [ ] Deploy and verify `curl "https://onboarding-service.tuleva.ee/v1/funds/EE0000003283/nav?format=csv"` before 16:00 Estonian returns data only up to 2 working days prior


src/main/java/ee/tuleva/onboarding/fund/FundController.java
```diff
@@ -42,8 +42,11 @@ public ResponseEntity<?> getNavHistory(
           .header(
               HttpHeaders.CONTENT_DISPOSITION,
               "attachment; filename=\"nav-" + isin.replaceAll("[^A-Za-z0-9]", "") + ".csv\"")
+          .header(HttpHeaders.CACHE_CONTROL, "no-store")
```
JordanValdma: I think we should manage this in API wide level, not in individual controllers. Use a pattern similar to ControllerAdvice or something that is more suitable

src/test/groovy/ee/tuleva/onboarding/fund/FundServiceSpec.groovy
JordanValdma, line 1: When you work with the groovy tests please re-write them into java unit tests, not to add or edit groovy code, which is essentially deprecated. Thanks!

src/test/groovy/ee/tuleva/onboarding/fund/FundControllerSpec.groovy
JordanValdma, line 1: When you work with the groovy tests please re-write them into java unit tests, not to add or edit groovy code, which is essentially deprecated. Thanks!

PR #1570, opened 2026-04-24
TKF: recurring payment link generator

stenehrlich-tuleva:
## Summary
- Backend URL-generator for Täiendav Kogumisfond recurring (standing-order) payments, mirroring the III pillar `RecurringPaymentLinkGenerator` pattern.
- Per-bank handling: LHV / Coop (COOP, COOP_WEB, PARTNER) / Swedbank build pre-fill standing-order URLs; SEB / Luminor return landing URLs; new `OTHER` returns `url=null` so the frontend renders a copy-card; `TULUNDUSUHISTU` explicitly throws as unsupported.
- Backend is the single source of truth for recipient name + IBAN (`payment-provider.savings-fund.*`). `PaymentLink` record extended with `recipientName` / `recipientIban` / `description` / `amount` (all nullable, `@JsonInclude(NON_NULL)`) so the frontend does not hardcode the IBAN.
- Swedbank URL pre-fills `account=<8-digit>` when the user's most recent `PROCESSED` deposit came from a Swedbank IBAN — new repo method `SavingFundPaymentRepository.findLastRemitterIban(PartyId)` via `JdbcClient`, ordered by `created_at DESC, id DESC` for idempotent output.
- Adds `PaymentType.SAVINGS_RECURRING` and `PaymentChannel.OTHER`; wires the new generator into `PaymentService`.

Issue: TulevaEE/TKF-VPII2026#18

Frontend counterpart lives on the onboarding-client branch `tkf-recurring-payment-links` and will be PR'd separately after this merges + deploys to staging, so the end-to-end flow can be verified against real data first.

## Deliberate trade-offs (flagged from an adversarial review pass)

1. **`PaymentServiceSpec.groovy` modified (existing Spock spec).** CI surfaced 9 failures because the new generator dep broke the positional `new PaymentService(...)` call. We only added the mock + constructor arg; no new test cases were added to the Spock file. Rewriting the full spec to Java is out of scope here.
2. **`RECURRING + OTHER` returns 500, not 400.** The new `OTHER` channel enlarges the enum globally, so the III pillar generator's exhaustive switch was updated to group `OTHER` with `TULUNDUSUHISTU` as unsupported (same `IllegalArgumentException`). Same pre-existing property `TULUNDUSUHISTU` has. A proper fix needs a cross-cutting `@ControllerAdvice` or per-type channel allowlist — left as follow-up.
3. **`OTHER` added to the shared `PaymentChannel` enum** rather than a narrower per-type union, to keep the single `GET /v1/payments/link` contract and avoid a dedicated endpoint. Plan-aligned trade-off.
4. **Spring Modulith boundary leak** — `payment/savings/recurring/` reads directly from `savings/fund/SavingFundPaymentRepository`. Consistent with the existing `SavingsCallbackService` pattern; tracked in TulevaEE/TKF-VPII2026#44.

## Test plan
- [x] New tests (all Java + JUnit 5, Spock avoided for new tests per PR #1565):
  - `PaymentControllerSavingsRecurringTest` — `@WebMvcTest` slice, delegation + JSON shape
  - `SavingsFundRecurringPaymentLinkGeneratorTest` — unit, every channel + edge cases (missing amount, malformed IBAN, reserved URL chars, Swedbank account= heuristic on/off)
  - `SavingsFundRecurringPaymentFlowIntegrationTest` — `@SpringBootTest` + `@Transactional`, end-to-end DB fixture verifying the Swedbank account= pre-fill
  - `SavingFundPaymentRepositoryTest` — 5 new cases for `findLastRemitterIban` (ordering, status filter, amount>0, party isolation, tied-timestamp idempotency)
- [x] CI green on `fbd203221`
- [ ] After merge + staging deploy: `curl "https://staging.tuleva.ee/v1/payments/link?type=SAVINGS_RECURRING&paymentChannel=LHV&amount=50&currency=EUR&recipientPersonalCode=<test>"` returns a valid LHV standing-order URL + the recipient fields
- [ ] Manual click-through of the returned Swedbank URL on staging: confirm recipient name renders as "Tuleva Täiendav Kogumisfond" (not `Tuleva+T%C3%A4iendav+...`)
- [ ] No DB migration — `remitter_iban` column on `saving_fund_payment` already exists


src/main/java/ee/tuleva/onboarding/payment/PaymentData.java
```diff
@@ -36,6 +37,7 @@ public enum PaymentChannel {
     COOP,
     COOP_WEB,
     PARTNER, // COOP_APP
-    TULUNDUSUHISTU
+    TULUNDUSUHISTU,
+    OTHER
```
vootelerotov: Hmm, adding OTHER as a value in enum is weird. Enum stands for `enumerate`, so things you can list. Adding OTHER there is weird, and I would argue, unexpected.

It is a known pattern in case of enums that are part of public contracts, to allow backwards compatibility of adding a new value, but here we do not use it like that.

Is there any other way we could proceed? Maybe turn Payment channel into optional parameter (what I would argue we are doing here)? Or add another endpoint, that does not require specifying payment channel.

src/main/java/ee/tuleva/onboarding/savings/fund/SavingFundPaymentRepository.java
```diff
@@ -100,6 +102,26 @@ public List<SavingFundPayment> findPaymentsWithStatus(PartyId partyId, Status...
         this::rowMapper);
   }
 
+  public Optional<String> findLastRemitterIban(PartyId partyId) {
```
ErkoRisthein: Why do we need this? Can't we just simplify and leave the iban empty for Swedbank? Helps us get rid of a lot of complexity in code with no clear downsides - Swedbank API params are not properly handled in Swedbank anyway (without an expliciity reload after login by the user).

src/main/java/ee/tuleva/onboarding/payment/PaymentDateProvider.java
```diff
@@ -27,7 +27,7 @@ LocalDate tenthDayOfMonth(LocalDate now) {
     return date;
   }
 
-  static String format(LocalDate date) {
+  public static String format(LocalDate date) {
```
ErkoRisthein, line 30: Why escalate the visibility to public here? Can't we keep it package-private, like designed? This class is only to be used inside the package. Maybe make the class package-private as well, to make the intent more explicit.

src/main/resources/application.yml
```diff
@@ -327,6 +330,9 @@ payment-provider:
     secret-key: ${PAYMENT_SECRET_SAVINGS}
     return-url: ${api.url}/payments/savings/callback
     notification-url: ${api.url}/payments/savings/notifications
+  savings-fund:
+    recipient-name: "Tuleva Täiendav Kogumisfond"
+    recipient-iban: "EE711010220306707220"
```
ErkoRisthein: This should go into environment variables. I.e. this property value should be set via env.

src/test/groovy/ee/tuleva/onboarding/payment/savings/recurring/SavingsFundRecurringPaymentFlowIntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.transaction.annotation.Transactional;
+
+@SpringBootTest
+@Transactional
+class SavingsFundRecurringPaymentFlowIntegrationTest {
```
ErkoRisthein: Why do we need an integration test for such a trivial flow (i.e. it just generates a link - a string)? Especially when we remove the `prefillsAccountFromLastProcessedRemitterIban` logic bloat.

Unit test should give us enough confidence for this simple use case. No need for a slow `@SpringBootTest`

src/test/groovy/ee/tuleva/onboarding/payment/savings/recurring/SavingsFundRecurringPaymentLinkGeneratorTest.java
```diff
@@ -0,0 +1,274 @@
...
+          recipientConfiguration, paymentDateProvider, savingFundPaymentRepository);
+
+  @BeforeEach
+  void setup() {
+    recipientConfiguration.setRecipientName(RECIPIENT_NAME);
+    recipientConfiguration.setRecipientIban(RECIPIENT_IBAN);
+    given(paymentDateProvider.tenthDayOfMonth()).willReturn(FIRST_PAYMENT);
+    given(
+            savingFundPaymentRepository.findLastRemitterIban(
+                new PartyId(PartyId.Type.PERSON, PERSONAL_CODE)))
```
ErkoRisthein: Static import PERSON

src/main/java/ee/tuleva/onboarding/payment/PaymentLink.java
```diff
@@ -1,5 +1,12 @@
...
-import jakarta.validation.constraints.NotNull;
+import com.fasterxml.jackson.annotation.JsonInclude;
 
-public record PaymentLink(@NotNull String url) {}
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record PaymentLink(
+    String url, String recipientName, String recipientIban, String description, String amount) {
+
+  public PaymentLink(String url) {
+    this(url, null, null, null, null);
```
ErkoRisthein: If you decided to make the backend the single source of truth for the IBAN, recipient name etc (which is good), then follow the boy-scout-rule and do the same for Pensionikeskus instead of making all the new fields nullable.

ErkoRisthein (changes requested):

ErkoRisthein:
Re-reviewed current head `1e393e6` plus the previous review threads. CI is green and the earlier Erko comments around Swedbank account prefill, env-backed TKF recipient config, removing the integration test, static import cleanup, and making Pensionikeskus data backend-owned look covered in code. The PaymentDateProvider thread is still unresolved in GitHub, but after the move to the payment module root I do not see the remaining public no-arg method as a merge blocker.

Not merge-ready yet: the savings-recurring amount validation still throws `IllegalArgumentException` from a controller path, which will surface as a 500 for client input, and the PR description/test plan is materially stale after later commits (still mentions `OTHER`, Swedbank `account=` prefill, `SavingFundPaymentRepository.findLastRemitterIban`, and the removed `@SpringBootTest`). Targeted local tests passed: `SavingsFundRecurringPaymentLinkGeneratorTest`, `RecurringPaymentLinkGeneratorSpec`, `PaymentLinkJacksonTest`.

src/main/java/ee/tuleva/onboarding/payment/savings/recurring/SavingsFundRecurringPaymentLinkGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+public class SavingsFundRecurringPaymentLinkGenerator implements PaymentLinkGenerator {
+
+  private static final BigDecimal MIN_AMOUNT = new BigDecimal("0.01");
+
+  private final SavingsFundRecipientConfiguration recipientConfiguration;
+  private final PaymentDateProvider paymentDateProvider;
+
+  @Override
+  public PaymentLink getPaymentLink(PaymentData paymentData, Person person) {
+    if (paymentData.getAmount() == null || paymentData.getAmount().compareTo(MIN_AMOUNT) < 0) {
```
ErkoRisthein: This is still a controller-reachable client validation path (`GET /v1/payments/link?type=SAVINGS_RECURRING`). `IllegalArgumentException` is not handled by `ErrorHandlingControllerAdvice`, so missing or too-small amounts return 500 instead of a structured 400. Please convert this to `ErrorsResponseException` (separate required/invalid codes would be ideal) and cover the endpoint behavior.

stenehrlich-tuleva: Fixed in 487fcf336.

- `IllegalArgumentException` → `ErrorsResponseException` so `ErrorHandlingControllerAdvice` maps to 400.
- Split into two stable codes: `payment.amount.required` (null amount, matching peer `RecurringPaymentLinkGenerator`) and `payment.amount.invalid` (below `MIN_AMOUNT` 0.01).
- Generator unit tests cover both codes via `.extracting(getCode)` (per CLAUDE.md "never assert on exception messages") plus an `acceptsExactMinimumAmount` boundary check at 0.01.
- `PaymentControllerSavingsRecurringTest` now exercises the real `SavingsFundRecurringPaymentLinkGenerator` via `willAnswer` (with mocked `PaymentDateProvider`/recipient config), so the endpoint behavior is covered end-to-end — a future revert to IAE would surface here as a 500 instead of being masked by a stub.

PR #1583, opened 2026-04-28
Use calculated AUM from nav_report for limit check denominator

taavipertman-glitch:
## Summary
- Position weight calculations now use the internally calculated AUM from `nav_report` UNITS row instead of `units × nav_per_unit`, eliminating small rounding differences caused by the rounded NAV per unit figure
- Falls back to `units × nav_per_unit` if nav_report is unavailable, throws `IllegalStateException` if neither source exists
- Wraps per-fund limit checks in try-catch so one fund's failure doesn't block the rest

## Test plan
- [x] Unit tests updated and passing (LimitCheckServiceTest)
- [x] Integration tests updated with nav_report UNITS rows and passing (LimitCheckIntegrationTest)
- [x] Full limit check test suite passing (47 tests)
- [x] Verify in staging that limit check results match the GS spreadsheet calculations

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
@@ -73,7 +73,9 @@ class LimitCheckService {
     }
 
     if (!errors.isEmpty()) {
-      var combined = new IllegalStateException("Limit check failed for %d fund(s)".formatted(errors.size()));
+      var combined =
+          new LimitCheckPartialFailureException(
+              "Limit check failed for %d fund(s)".formatted(errors.size()), results);
       errors.forEach(combined::addSuppressed);
       throw combined;
     }
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
@@ -53,6 +53,7 @@ class LimitCheckService {
 
   List<LimitCheckResult> runChecksAsOf(LocalDate asOfDate) {
     var results = new ArrayList<LimitCheckResult>();
+    var errors = new ArrayList<Exception>();
 
     for (var fund : TulevaFund.values()) {
       var latestDate = fundPositionRepository.findLatestNavDateByFundAndAsOfDate(fund, asOfDate);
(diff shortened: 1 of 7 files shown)


src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
```diff
@@ -62,8 +62,12 @@ List<LimitCheckResult> runChecksAsOf(LocalDate asOfDate) {
       }
 
       var checkDate = latestDate.get();
-      var result = checkFund(fund, checkDate);
-      results.add(result);
+      try {
+        var result = checkFund(fund, checkDate);
+        results.add(result);
+      } catch (Exception e) {
```
JordanValdma, line 69: Whole pipeline should already have error reporting to slack. Doesn't this new line eat the errors and break the slack messages pipeline?

src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
```diff
@@ -151,6 +155,11 @@ private BigDecimal signedCashImpact(TransactionOrder order) {
   }
 
   private BigDecimal computeTotalNav(TulevaFund fund, LocalDate checkDate) {
+    var calculatedAum = navReportPositionProvider.getCalculatedAum(fund, checkDate);
+    if (calculatedAum.isPresent()) {
+      return calculatedAum.get();
+    }
+    log.warn("Calculated AUM unavailable in nav_report, falling back to units × NAV per unit: fund={}", fund);
```
JordanValdma: Should it be warn or error?

JordanValdma: Does it need a slack message?

JordanValdma (approved):

src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
taavipertman-glitch: No — the per-fund try/catch collects errors and throws `LimitCheckPartialFailureException` after all funds are attempted. `LimitCheckJob` catches it, calls `pipelineTracker.stepFailed()`, and still notifies on any breaches found. So Slack pipeline reporting is preserved. Tests added: `partialFailureNotifiesBreachesAndMarksStepFailed` and `partialFailureWithoutBreachesSkipsNotification`.

src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
taavipertman-glitch: Changed to `log.warn` — it's a successful fallback, not an error. No separate Slack message needed since the pipeline tracker already reports step failures if the fallback also fails.

PR #1594, opened 2026-05-05
Handle in-runoff instruments in health checks and remove NAV publish gate

taavipertman-glitch:
## Summary

- When model portfolio is updated (instrument removed/replaced), positions still holding the old instrument no longer block import with FAIL
- `IsinMatchChecker` now checks both current and previous model portfolio allocations
- ISIN in previous but not current model → WARNING (in-runoff, awaiting sell)
- ISIN in neither → FAIL (truly unknown instrument)

## Changes

- `ModelPortfolioAllocationRepository`: added `findPreviousByFund()` query (second-to-latest effective_date)
- `IsinMatchChecker`: accepts previous allocations, distinguishes in-runoff (WARNING) from unknown (FAIL)
- `HealthCheckService`: fetches and passes previous allocations

## Test plan

- [x] `IsinMatchCheckerTest` — in-runoff → WARNING, unknown → FAIL, current → PASS
- [x] `ModelPortfolioAllocationRepositoryTest` — `findPreviousByFund` with multiple versions, single version, no data
- [x] `HealthCheckServiceTest` — updated mocks pass
- [ ] Deploy to staging, update model portfolio, verify old ISINs produce WARNING not FAIL

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1595, opened 2026-05-05
Handle in-runoff instruments across tracking, limits, and transactions

taavipertman-glitch:
## Summary

Extends the in-runoff instrument handling from #1594 to four more components that use model portfolio allocations. During model portfolio transitions (1-7 days between model change and trade settlement), these components previously caused false alarms or silent gaps.

- **TrackingDifferenceService**: in-runoff securities now included with `modelWeight=0`, preventing false breach alerts and inflated residuals. Missing prices for in-runoff securities no longer throw `IncompletePriceDataException`.
- **LimitCheckService**: `buildIsinToProviderMap` merges previous + current allocations so in-runoff positions count toward provider concentration.
- **TransactionInputService**: `getInstrumentTypes` and `getOrderVenues` merge previous + current so sell orders for old instruments get correct type/venue instead of defaults.
- **TransactionPreparationService**: label/ticker/bbgTicker lookups merge previous + current so SEB Fund, SEB ETF, and FT ETF export files have complete data for in-runoff sell orders.

## Test plan

- [x] `TrackingDifferenceServiceTest` — in-runoff securities appear with zero model weight and correct actual weight
- [x] `TrackingDifferenceServiceTest` — missing prices for in-runoff securities don't throw
- [x] Existing tracking, limit, transaction, and health check tests pass
- [x] Full test suite green
- [ ] Deploy to staging, change model portfolio, verify no false tracking difference breaches

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
@@ -19,6 +19,7 @@ import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.ZoneOffset;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


JordanValdma (approved):
Reviewed via agent-assisted code review. Propagates in-runoff awareness across TrackingDifferenceService, LimitCheckService, TransactionInputService, TransactionPreparationService with consistent 'current wins on collision' semantics. Follow-up: add a direct merge test for TransactionPreparationService (current tests pass only because Mockito returns empty by default), and document BENCHMARK_MODEL TD inclusion of in-runoff securities.

PR #1596, opened 2026-05-06
TKF fall back to remitter id code when description lacks code

stenehrlich-tuleva:
## Summary

If a Täiendava Kogumisfondi payment arrives without a personal/registry code in the description but the bank statement carries a valid id in the remitter id field (`RltdPties/Dbtr/Id/PrvtId/Othr/Id` from camt.052/053), use that instead of returning the money.

The existing cross-check between description and remitter id still rejects mismatches (e.g. selgituses olev kood ≠ maksja isikukood), so this only changes behavior for the slice where the description gives us nothing but the bank already told us who the payer is. `parsePartyId` (strict, no substring extraction) is used for the remitter id field, matching how it's used elsewhere — so Wise customer ids like `P13694547` still don't match.

## Test plan

- [x] `process_success_fallsBackToRemitterIdCodeWhenDescriptionLacksCode` — new unit test for the fallback (description has no code, remitter id code matches an onboarded user → VERIFIED + recordPaymentReceived)
- [x] `process_returnsPaymentWhenNoCodeAnywhere` — renamed/repurposed from old `process_noCodeInDescription`; covers the genuine "no code anywhere" case (still TO_BE_RETURNED)
- [x] `process_personalCodeMismatch` — passes unchanged (cross-check intact)
- [x] All 21 `PaymentVerificationServiceTest` cases green
- [x] `savings.fund.*` + `banking.processor.*` suites green locally
- [x] CircleCI workflow green on this branch

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
+++ b/src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
@@ -68,16 +68,10 @@ public class PaymentVerificationService {
   public void process(SavingFundPayment payment) {
     log.info("Processing payment {}", payment.getId());
 
-    var partyIdFromDescription = extractPartyId(payment.getDescription());
-    var partyIdOpt = partyIdFromDescription.or(() -> parsePartyId(payment.getRemitterIdCode()));
+    var partyIdOpt = extractPartyId(payment);
     if (partyIdOpt.isEmpty()) {
-      identityCheckFailure(payment, "selgituses puudub isikukood/registrikood");
+      identityCheckFailure(payment, "makse ei sisalda tuvastatavat isikukoodi/registrikoodi");
       return;
-    }
-    if (partyIdFromDescription.isEmpty()) {
-      log.info(
-          "Payment {} has no code in description, falling back to remitter id code",
-          payment.getId());
     }
 
     PartyId partyId = partyIdOpt.get();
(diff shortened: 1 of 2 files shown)


src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
```diff
@@ -68,13 +68,19 @@ static VerificationMessages forType(PartyId.Type type) {
   public void process(SavingFundPayment payment) {
     log.info("Processing payment {}", payment.getId());
 
-    var extractPartyId = extractPartyId(payment.getDescription());
-    if (extractPartyId.isEmpty()) {
+    var partyIdFromDescription = extractPartyId(payment.getDescription());
+    var partyIdOpt = partyIdFromDescription.or(() -> parsePartyId(payment.getRemitterIdCode()));
+    if (partyIdOpt.isEmpty()) {
       identityCheckFailure(payment, "selgituses puudub isikukood/registrikood");
```
vootelerotov: This is not fully correct any more, night -- because this means that there is no code in the description AND we did not manage to parse it from the remitter ID code.

vootelerotov: "makse ei sisalda tuvastatavat isikukoodi/registrikoodi"?

src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
```diff
@@ -68,13 +68,19 @@ static VerificationMessages forType(PartyId.Type type) {
...
       return;
     }
+    if (partyIdFromDescription.isEmpty()) {
+      log.info(
+          "Payment {} has no code in description, falling back to remitter id code",
+          payment.getId());
+    }
```
vootelerotov: I would suggest moving this to a separate method, called `extractPartyId`, and renaming the current `extractPartyId` to `extractPartyIdFromDescription`:.

Then you can do something like this:

```
    var partyIdFromDescription = extractPartyIdFromDescription(payment.getDescription());

   if (partyIdFromDescription.iPresent()) {
     return partyIdFromDescription;
   }

    log.info( "Payment {} has no code in description, falling back to remitter id code",  payment.getId());

   return parsePartyId(payment.getRemitterIdCode())
```

The same thing happens at the moment, but the check if we had it from description happens after the check if we have it from either source, what is a bit weird.

vootelerotov (approved):
Two comments. I would fix the error message, as it is not correct any more. The idea of extracting the logic into a method is not functional, but I do think it improves readability.

Approving with assumption that the error message is improved before merging.

src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
stenehrlich-tuleva: Hea soovitus, võtsin täpselt selle teksti — fixed.

src/main/java/ee/tuleva/onboarding/savings/fund/PaymentVerificationService.java
stenehrlich-tuleva: Tehtud — vana `extractPartyId(String)` on nüüd `extractPartyIdFromDescription`, ja uus private `extractPartyId(SavingFundPayment)` katab fallback'i + log-i. `process()` on nüüd üks rida selle asemel.

PR #1600, opened 2026-05-07
Throw on rejected/invalid Mandrill response in sendSystemEmail

taavipertman-glitch:
## Summary
- `sendSystemEmail()` now checks `MandrillMessageStatus.getStatus()` after the API call
- If status is "rejected" or "invalid", throws `EmailDeliveryException` so callers see the failure instead of silently treating it as success
- Without this, a rejected email would cause rows to be marked as published even though nothing was delivered

## Test plan
- [x] New Spock tests: `sendSystemEmail throws when Mandrill returns rejected` and `...invalid`
- [x] Existing `EmailServiceSpec` tests still pass
- [x] Caller tests (`NavReportEmailSenderTest`, `TrusteeReportJobTest`) still pass

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/notification/email/EmailDeliveryException.java
+++ b/src/main/java/ee/tuleva/onboarding/notification/email/EmailDeliveryException.java
@@ -1,0 +1,8 @@
+package ee.tuleva.onboarding.notification.email;
+
+public class EmailDeliveryException extends RuntimeException {
+
+  public EmailDeliveryException(String message) {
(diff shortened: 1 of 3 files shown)


JordanValdma (approved):
Reviewed via agent-assisted code review. Throws on rejected/invalid Mandrill response; fixes silent-failure bug where NAV report job marked rows as published despite Mandrill rejection. Minor duplication between send() and sendSystemEmail() — acceptable follow-up.

PR #1603, opened 2026-05-07
Remove FREE_CASH breach from Slack notifications

taavipertman-glitch:
## Summary
- Remove FREE_CASH breach from Slack limit check notifications — the 1M free cash threshold was generating noise
- The free cash limit check still runs and events are persisted to the database for audit
- `hasBreaches()` no longer considers `freeCashBreach`, so a FREE_CASH-only breach results in "within limits"

## Test plan
- [ ] Verify limit check still runs and `FREE_CASH` events are saved in `limit_check_event` table
- [ ] Verify Slack notification no longer includes FREE_CASH lines
- [ ] Verify other breach types (POSITION, PROVIDER, RESERVE) still notify correctly

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckNotifier.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckNotifier.java
@@ -78,14 +78,6 @@ class LimitCheckNotifier {
                       breach.reserveSoft(),
                       breach.reserveHard()));
         }
-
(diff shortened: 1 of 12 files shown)


JordanValdma (approved):
Reviewed via agent-assisted code review. FREE_CASH Slack alerts removed; DB audit trail (LimitCheckEvent FREE_CASH rows) preserved. NAV pipelines split per fund. Added a regression test (97a30779a) locking in the getCalculatedAum published_at scoping — the sibling getSecurityMarketValues was already doing this filter, so this brings the two methods in line.

PR #1618, opened 2026-05-13
Use calculation date in NAV CSV filename

taavipertman-glitch:
## Summary
- NAV CSV attachment filename now uses `calculationDate` (when NAV was calculated) instead of `positionReportDate` (the NAV date)
- Email subject still uses the NAV date, which is correct
- Example: for May 7 NAV calculated on May 8, subject = "TKF100 NAV arvutamine 07.05.2026", filename = "TKF100 NAV arvutamine 08052026.csv"

## Why
Consumer scripts (limit_check, tracking_difference, etc.) identify CSV files by the filename date, which historically matched the calculation date. The onboarding-service was using the position report date, causing filename mismatches.

## Test plan
- [x] `NavReportEmailSenderTest` updated and passing

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/savings/fund/nav/NavReportEmailSender.java
+++ b/src/main/java/ee/tuleva/onboarding/savings/fund/nav/NavReportEmailSender.java
@@ -33,6 +33,7 @@ class NavReportEmailSender {
 
     var fundCode = result.fund().getCode();
     var navDate = result.positionReportDate();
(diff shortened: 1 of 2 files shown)


JordanValdma (approved):
Reviewed: filename now reflects when the calc was done rather than the as-of position date — sensible when Friday positions are NAV'd on Monday. Test asserts the resulting filename. Clean.

PR #1619, opened 2026-05-13
Add investment report publishing pipeline

taavipertman-glitch:
## Summary

Port the investeeringute aruanded process from GAS into onboarding-service. Full pipeline: DB → PDF → WordPress → email.

- **Data layer:** Query `nav_report` + `instrument_reference` + `investment_portfolio_cost_basis` tables, calculate NAV percentages, month-over-month changes, and avg cost from transaction registry
- **PDF generation:** Thymeleaf HTML template → openhtmltopdf, matching the existing GAS report layout
- **WordPress upload + ACF update:** Upload PDFs to media library via REST API, then update `investment_report_file` ACF field on each fund page — report URLs go live immediately, no git commits needed
- **Email:** Send via Mandrill (existing `EmailService`) with 3 PDF attachments — same pattern as NAV report emails. Initially sends to `funds@tuleva.ee` only for internal review
- **Admin endpoint:** `POST /admin/publish-investment-reports?month=3&year=2026`
- **Preview endpoint:** `GET /admin/preview-investment-report?fundCode=TUK75&month=3&year=2026` — returns PDF for visual comparison, no credentials needed
- **Cost basis integration:** Avg cost per unit and total cost populated from `PortfolioCostBasisService` (transaction registry)
- **Quantity validation:** Pre-publish warning when NAV report quantities don't match cost basis ledger (catches late settlements)

### Credentials needed

Only one new credential: **WordPress Application Password** (`WP_USERNAME` + `WP_APP_PASSWORD`). Used for both PDF upload and ACF field updates. Email uses existing Mandrill.

### One-time WordPress theme setup (separate PR)

- Add `investment_report_file` ACF field to TUK75, TUK00, TUV100 fund pages
- Update 3 PHP templates to read from `get_field('investment_report_file')` instead of hardcoded URLs

### Pre-publish validations

- All 4 funds have published NAV data for the month
- All funds share the same NAV date
- NAV date is end-of-month (day >= 25)
- Total asset allocation ~100% per fund (±1.5% tolerance)
- Quantity reconciliation: NAV report vs cost basis ledger (warning, non-blocking)

## Test plan

- [x] Unit tests for data service (10 tests) — percentages, fallbacks, previous month changes, cost basis integration, quantity validation
- [x] Unit tests for PDF generator (3 tests) — PDF bytes, HTML content, placeholder rendering
- [x] Unit tests for orchestrator (9 tests) — full pipeline, TKF100 exclusion, partial failure, NAV date/allocation validation, email send, ACF update
- [x] Controller tests for publish and preview endpoints
- [ ] Visual PDF comparison with GAS output (use preview endpoint)
- [ ] End-to-end test with real WordPress credentials


JordanValdma:
Pushed a commit (`4cd41095b`) fixing the confirmed defects from a review pass. Full suite + coverage + spotless green. Summary:

**Fixed**
- **All-or-nothing publish** — upload every PDF first and abort before touching any fund page if a PDF or upload fails, so one failure can't leave some fund pages on the new month and others on last month's. A page-update failure *after* all uploads succeed is treated as a genuine partial publish and now sends an alert email; a fund's URL is recorded only after its page is actually repointed.
- **Fail-fast on missing data** (per CLAUDE.md) — a `SECURITY` NAV row with no market value now throws instead of reporting the holding at €0; the securities **total cost is left blank** when any holding lacks a cost-basis snapshot instead of emitting a misleading partial sum (the old `treatsNullSecurityMarketValueAsZero` test is inverted).
- **WordPress RestClient timeouts** (5s connect / 30s read) — matches `PopulationRegisterConfiguration`/`SebGatewayConfiguration`; a stalled WP no longer hangs the request thread.
- **Media dedup now works** — `sanitizeFilename` produces the exact WordPress slug (lowercase, ASCII-folded, hyphenated), so the reuse check matches the stored `source_url` instead of re-uploading a duplicate every run. (Previously it only stripped `\"`/`\r`/`\n`, so the spaces+capitals in the real filename never matched.)

**Left for you (product calls, not fixed)**
- **Same-NAV-date across all 4 funds incl. TKF100** is an exact `LocalDate` equality check — a 1-day TKF100 skew aborts the 3 pension-fund reports. Worth confirming all four truly share a NAV calendar, or validating per publish cohort.
- **Allocation ratio numerator** (`totalAssetsMarketValue`) excludes unsettled-transaction receivables/payables, but `fundNav` is net of them — if those rows are material the ±1.5% gate can false-trip. Worth confirming they net to ~0.
- Minors: blank WP creds don't fail fast; no email attachment-size guard; quantity reconciliation is one-directional. Also the PR description says preview needs "no credentials" — it actually does require `X-Admin-Token` (good — just a doc nit).

Happy to take any of the "left for you" items if you'd like.

JordanValdma:
Verified the two open questions against **live production data** + the **GAS source**, and pushed the one real remaining fix.

**Same-NAV-date — confirmed safe, no change.** Every month-end for the last 3 months has all 4 funds (`TUK00/TUK75/TUV100/TKF100`) publishing on the *exact same* date, so the strict same-date check never false-aborts. Leaving it as-is.

**Allocation numerator (assets ~100.02% vs NAV) — not a gap, matches GAS by design.** The dropped rows are `LIABILITY` (payables) + `LIABILITY_FEE` (accrued mgmt/custody fee) — the management fee is the only material one (e.g. €177k on TUK75 = 0.016% of NAV). I checked the legacy GAS report (`investeeringute_aruanne.gs`): **it deliberately omits all liabilities/fees and shows gross assets at ~100.x%**, with NAV printed separately and the small gap absorbed by the ±1.5% tolerance. This Java port already reproduces that exactly. Live ratios are 100.011%–100.020% — ~75× inside the band. Adding a liabilities line would *diverge* from GAS and fail the visual comparison, so we should **not** change it. (SQL used is in the thread.)

**Pushed: WordPress `RetryTemplate` (`c23d064b4`).** The one genuine remaining merge-readiness item — wraps the 4 WP calls in the project-standard Spring 7 retry (5xx + IO, never 4xx, 2 retries / 500ms backoff), now safe because the slug-dedup made upload idempotent. New test covers retry-then-succeed.

Net: both questions resolved, retry added. From my side this is merge-ready; the only remaining checkbox is the operator visual-PDF comparison.

PR #1643, opened 2026-05-25
Add public investor-count endpoint + guardrail for tuleva.ee front page

stenehrlich-tuleva:
## Summary

Automates the "85 224 Eesti inimest kogub Tuleva indeksfondides" figure on the tuleva.ee front page, which is currently edited by hand in wp-admin. Adds a public endpoint that returns the count of active investors, read from the same materialized view our Metabase dashboard uses.

## What's included

- **`GET /v1/statistics/investor-count`** → `{"count": N}` — public (`permitAll`), `Cache-Control: public, max-age=1h`.
- Reads `total_active_investors` at the latest `reporting_date` from `analytics.mv_kpi_new` (II + III pillar, deduplicated — the same figure as the Metabase "kogujate arv"). The aggregate query always returns exactly one row and never throws on an empty/NULL view.
- New `statistics` module: thin controller → `InvestorStatisticsService` → `InvestorStatisticsRepository`.
- **Guardrail**: `InvestorCountGuardrailJob` (daily 07:30 Tallinn, production/staging) reads the current + previous-period count and logs an `ERROR` if the count leaves configurable bounds **or** jumps more than a configurable % vs the previous reporting period. The Sentry logback appender forwards that ERROR to **#tech** (no new Slack webhook). Tunable via `investor-count.guardrail` (defaults: 80k–95k band, 5% max change). The period-over-period check is the primary anomaly catcher; the band may need bumping as the investor count grows past 95k.

## Not included (deferred)

- **TKF (savings-fund) clients**, including legal entities (juriidilised isikud) — to be added once the combined Metabase definition is settled with Marju; the endpoint will then mirror that SQL.

## Test plan

- [x] Unit + slice tests: controller public-access via the real `SecurityConfiguration`, repository SQL, guardrail logic (bounds / period jump / no-previous / jump-from-zero), job alerting, service
- [x] CircleCI green
- [x] codex review clean
- [ ] After deploy: `curl https://onboarding-service.tuleva.ee/v1/statistics/investor-count` returns a count matching the latest Metabase `total_active_investors`


src/main/resources/application.yml
```diff
@@ -192,6 +192,12 @@ transaction-registry:
     cc:
       - taavi.pertman@tuleva.ee
 
+investor-count:
```
JordanValdma: This seems like code that needs to be again updated in the future. I would take it away if possible

src/main/java/ee/tuleva/onboarding/statistics/InvestorCountGuardrail.java
JordanValdma, line 1: I think is sub 0.1% per day probability that this code is ever needed

JordanValdma: Along with all of the guardrails

src/main/java/ee/tuleva/onboarding/statistics/InvestorStatisticsRepository.java
```diff
@@ -0,0 +1,42 @@
...
+@RequiredArgsConstructor
+public class InvestorStatisticsRepository {
+
+  private final JdbcClient jdbcClient;
+
+  public long getActiveInvestorCount() {
+    return jdbcClient
+        .sql(
+            """
+            SELECT COALESCE(SUM(total_active_investors), 0)
```
JordanValdma, line 17: If you want to add any guardrails then i would do those in the sql, or a dedicated sql view

JordanValdma (changes requested):

JordanValdma (approved):
Guardrail machinery removed; replaced with an inline HAVING sanity bound (50k–500k) on the investor-count query. Requested changes addressed.

PR #1646, opened 2026-05-26
Tkf see if we can improve oom in ci

vootelerotov:


src/test/java/org/springframework/core/type/classreading/MetadataReaderFactoryDelegate.java
```diff
@@ -0,0 +1,23 @@
...
+// Test-only shadow of Spring's package-private MetadataReaderFactoryDelegate. It forces the
+// ASM-based SimpleMetadataReaderFactory instead of the JDK ClassFile API reader that Spring
+// Framework 7 selects on JDK 24+. The ClassFile reader pins parsed class readers (raw bytes +
+// constant pool) via bound annotations retained in bean definitions / per-context resource
+// caches, which blows up heap when many Spring test contexts are cached and OOMs CI.
+// See https://github.com/spring-projects/spring-framework/issues/36737.
+//
+// Lives on the test classpath only (loaded before spring-core.jar), so the production artifact
+// keeps Spring's default reader. Remove once spring-framework#36737 is resolved.
+abstract class MetadataReaderFactoryDelegate {
```
vootelerotov, line 14: FYI @JordanValdma and @ErkoRisthein  -- this is quite a hack (see package name), but the issue does not seem to have a better workaround on Spring side.

I got OOM without this every 2nd test run, started losing my sanity.

As it is a hack, I would want more people to be aware of it, just in case :)

No action needed!

PR #1647, opened 2026-05-26
Serve TKF onboarding document links from WordPress (GET /v1/savings/documents)

stenehrlich-tuleva:
## What & why

The Täiendav Kogumisfond (TKF100) onboarding **signing step** links three legal documents — **terms (tingimused), prospectus (prospekt), key information (põhiteave)**. Those URLs were **hardcoded in the frontend** and silently went stale when the documents were re-issued.

This adds **`GET /v1/savings/documents`**, which mirrors the current URLs from the public fund-documents page's **WordPress ACF fields** and serves them to the frontend. WordPress stays the single editable source — the content team updates the documents in wp-admin, **no deploy anywhere**.

Issue: TulevaEE/TKF-VPII2026#63

## Hard requirement: a WordPress outage must never fail a deploy or break the signing page

Solved by a baked-in fallback + zero-network startup + graceful scheduled refresh:

- **In-memory `AtomicReference`** seeded with a **baked-in fallback** (today's effective URLs, verified 2026-05-26). Not `@Cacheable` — that would fetch on the request path on a miss.
- **Startup performs zero network I/O.** No bean-init/`ApplicationReadyEvent` fetch; config has safe defaults baked into the class. WordPress is contacted **only** by an hourly `@Scheduled` refresh (`production`/`staging`, `@SchedulerLock`).
- `refresh()` **never propagates**: on failure it keeps last-known-good, alerts operations (the alert send is itself guarded), and escalates a **staleness alert after 48h** on the fallback so a permanently-failing fetch can't sit silently.
- The endpoint is **`permitAll`** so an auth state can't blank the signing step.

### The key proof
`SavingsFundDocumentsIntegrationTest` is a context-level `@SpringBootTest` wired with a WordPress client that **always fails**, asserting the app **boots**, serves the **fallback** to an **unauthenticated** request, and **never calls WordPress** on the startup/request path.

## Design notes (house patterns)
- **Anti-corruption layer**: the WordPress JSON DTOs (`WordpressPage`/`Acf`) are **package-private**; validated and mapped to the public `SavingsFundDocuments` record at the boundary. URL validation rejects anything that isn't `https`, host `tuleva.ee`, path under `/wp-content/uploads/`, ending `.pdf`, no query/fragment. (`ModularityTest.verify()` is disabled, so visibility is enforced by hand.)
- **Native `RetryTemplate`** (Spring FW 7) retries 5xx / `ResourceAccessException`, not 4xx — mirrors `SebGatewayConfiguration`.
- **`Clock` injected**; `@ConfigurationProperties(prefix = "wordpress-documents")` with defaults baked into the class so missing config can't fail startup.
- **Alerting via `OperationsNotificationService` (SAVINGS channel)** rather than building a Mandrill message directly — that's the public ops-alert API and avoids reaching into another module's package-private `AlertMandrillMessageFactory`/`AlertProperties`. The "alert must not propagate" requirement is fully met (and the integration/unit tests prove `refresh()` never throws even when the alert send throws).

## Reviewer flags
- 👀 **Security**: please glance at the new `permitAll` for `GET /v1/savings/documents` — it exposes only URLs that are already public on the website.
- ⏳ **Live E2E waits on the WordPress prereq**: `show_in_rest` for the ACF group (TulevaEE/wordpress-theme#67) is likely **not deployed yet**, so live WordPress REST won't expose the fields. This PR is built and tested against the documented contract with mocked HTTP; the **baked-in fallback makes the app correct regardless**. Once #67 deploys, verify with:
  ```bash
  curl -s 'https://tuleva.ee/wp-json/wp/v2/pages?slug=tuleva-taiendav-kogumisfond-dokumendid' \
    | python3 -c 'import sys,json;a=json.load(sys.stdin)[0]["acf"];print(a["terms_file"],a["prospectus_file"],a["key_investor_info_file"],sep="\n")'
  ```

## Tests
Strict TDD. Controller (`@WebMvcTest`), anti-corruption mapping + retry (`MockRestServiceServer`), resilience + staleness (unit), and the context-level deploy-safety `@SpringBootTest`. `./gradlew spotlessApply` clean; **full `./gradlew test` suite green**.

## Frontend
Reads this endpoint: TulevaEE/onboarding-client#1557

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/config/SecurityConfiguration.java
+++ b/src/main/java/ee/tuleva/onboarding/config/SecurityConfiguration.java
@@ -56,6 +56,8 @@ public class SecurityConfiguration {
                     .permitAll()
                     .requestMatchers(GET, "/v1/funds/*/nav")
                     .permitAll()
+                    .requestMatchers(GET, "/v1/savings/documents")
+                    .permitAll()
                     .requestMatchers(HEAD, "/v1/members")
                     .permitAll()
                     .requestMatchers(GET, "/v1/members/lookup")
--- a/src/main/java/ee/tuleva/onboarding/savings/fund/documents/SavingsFundDocuments.java
+++ b/src/main/java/ee/tuleva/onboarding/savings/fund/documents/SavingsFundDocuments.java
@@ -1,0 +1,3 @@
+package ee.tuleva.onboarding.savings.fund.documents;
+
+public record SavingsFundDocuments(String terms, String prospectus, String keyInformation) {}
--- a/src/main/java/ee/tuleva/onboarding/savings/fund/documents/SavingsFundDocumentsController.java
+++ b/src/main/java/ee/tuleva/onboarding/savings/fund/documents/SavingsFundDocumentsController.java
@@ -1,0 +1,21 @@
+package ee.tuleva.onboarding.savings.fund.documents;
+
+import io.swagger.v3.oas.annotations.Operation;
+import lombok.RequiredArgsConstructor;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/v1/savings")
+@RequiredArgsConstructor
+public class SavingsFundDocumentsController {
+
+  private final SavingsFundDocumentsService savingsFundDocumentsService;
+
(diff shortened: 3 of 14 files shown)


stenehrlich-tuleva:
Parking the WordPress-backed approach for now — it's more than we want to take on before our CTO is back to review the design. The branch stays put for pickup; tracking in TulevaEE/TKF-VPII2026#63. In the meantime we'll ship a minimal interim fix that just bumps the hardcoded document URLs to the current in-force (Feb 2026) versions.

PR #1648, opened 2026-05-26
Tkf allow withdrawls only from bank account with matching remitter

vootelerotov:


src/main/java/ee/tuleva/onboarding/savings/fund/SavingFundPaymentRepository.java
```diff
@@ -100,12 +100,13 @@ public List<SavingFundPayment> findPaymentsWithStatus(PartyId partyId, Status...
...
   }
 
-  public List<String> findDepositBankAccountIbans(PartyId partyId) {
+  public List<String> findWithdrawableIbans(PartyId partyId) {
     return jdbcTemplate.query(
         """
             SELECT DISTINCT remitter_iban
             FROM saving_fund_payment
             WHERE party_type = :party_type AND party_code = :party_code
+              AND remitter_id_code = :party_code
```
vootelerotov, line 109: @stenehrlich-tuleva, does not seem too narrow to you?

Note from Claude:

`Company remitters use OrgId, not PrvtId — and the extractor reads only getPrvtId() (line 76-84), ignoring organisation IDs. So a deposit from a business account gets remitter_id_code = null.`

Should double-check this and if it is the case, fix before go-live.

PR #1650, opened 2026-05-27
TKF #67: lock KYC short-survey → identical onboarding outcome (regression test)

stenehrlich-tuleva:
## What

Adds `KycShortSurveyRegressionTest` — a `@SpringBootTest` integration test that locks the assumption the [TKF company-onboarding (#67)](https://github.com/TulevaEE/TKF-VPII2026/issues/67) **frontend-only** approach depends on:

> A KYC survey that omits the profile items (`INVESTMENT_GOALS`, `INVESTABLE_ASSETS`, `SOURCE_OF_INCOME`) **and** the personal `TERMS` item is accepted and drives PERSON onboarding to the **same `COMPLETED` outcome** as the full survey.

This is the "company-only person = no profile" invariant Codex flagged: it must be test-locked, not optional. The whole frontend-only plan rests on the backend not requiring profile answers.

## Why

The frontend will let a "only via company" user skip the profile + personal-terms steps and submit a shorter `POST /v1/kyc/surveys` payload. This test proves the backend already accepts that and completes PERSON onboarding exactly as today — so no backend change is needed, and the contract can't silently regress.

## How

- Posts a **short** survey (citizenship, address, email, phone, PEP — no profile, no terms) → asserts `200`, survey persisted (5 answers), `KycCheckPerformedEvent` fired, PERSON status `COMPLETED`.
- Posts a **full** survey → asserts the identical `COMPLETED` outcome.

**No production code change.** Confirmed green locally without touching prod code, which is the expected result (the assumption holds).

## Honest caveat (in the test's Javadoc)

The risk level itself is produced by the external `kyc_ob_assess_user_risk` SQL function (deployed from `database-administration`, not in this service's schema), so it's stubbed by `TestKycChecker`. The profile fields' irrelevance to the *risk score* is a property of that external function and out of scope here. What this test guarantees is the **service-side contract**: no validation requires the profile items and the onboarding-completion path is agnostic to them.

### DIFF
--- a/src/test/groovy/ee/tuleva/onboarding/kyc/survey/KycShortSurveyRegressionTest.java
+++ b/src/test/groovy/ee/tuleva/onboarding/kyc/survey/KycShortSurveyRegressionTest.java
@@ -1,0 +1,157 @@
+package ee.tuleva.onboarding.kyc.survey;
+
+import static ee.tuleva.onboarding.auth.AuthenticatedPersonFixture.authenticatedPersonFromUser;
+import static ee.tuleva.onboarding.auth.UserFixture.sampleUserNonMember;
+import static ee.tuleva.onboarding.auth.authority.Authority.USER;
+import static ee.tuleva.onboarding.party.PartyId.Type.PERSON;
+import static ee.tuleva.onboarding.savings.fund.SavingsFundOnboardingStatus.COMPLETED;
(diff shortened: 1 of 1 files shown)


vootelerotov:
Looks good to me!

vootelerotov (approved):

PR #1691, opened 2026-06-11
AML(#78) M1: attribute KYB checks to company via aml_check.company_id

stenehrlich-tuleva:
**Milestone M1** of the company AML risk-scoring rollout (issue TulevaEE/tuleva#78, Phase A — data capture). Self-contained spec: the rev-3 comment on tuleva#78 and the plan file in TulevaEE/tuleva#79.

## Problem
KYB checks are attributed to a company only via the representative's `personal_code` (`aml_check.personal_code → company_party.party_code`). A person who represents two OÜs contaminates both companies' scores — the Codex rev-3 **P1 ship-blocker** and locked decision #1 in the plan.

## Change
- New column `aml_check.company_id uuid` (FK `company(id)`, indexed) — migration `V1_207`.
- `AmlKybCheckEventListener` resolves `company_id` once per KYB event from the event's registry code (`CompanyRepository.findByRegistryCode`) and stamps it on every check.
- Non-KYB (person/KYC) checks leave `company_id` null. A KYB check whose `Company` row isn't yet persisted also leaves it null — by design, so M7 treats it as **NO_DATA** rather than mis-attributing by representative.

## Tests (TDD)
- `AmlKybCheckEventListenerTest`: resolves company_id; **cross-company isolation** (rep in two companies → check carries only the event company); null-when-not-persisted.
- `AmlServiceIntegrationTest` green in isolation (boots Flyway → V1_207 applies on H2; entity↔schema consistent). `spotlessCheck` clean.

## Notes
- Foundation for M7's `analytics.v_company_risk_metadata` matview, which will key `kyb_latest` on `company_id` (no `personal_code` fallback).
- `V1_207` is the next free slot after master's `V1_206`; rebase just before merge if another migration lands first.
- `company_id` is intentionally **nullable** (most `aml_check` rows are person/KYC checks), so no `SET NOT NULL`.

Refs TulevaEE/tuleva#78.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/aml/AmlCheck.java
+++ b/src/main/java/ee/tuleva/onboarding/aml/AmlCheck.java
@@ -9,6 +9,7 @@ import jakarta.validation.constraints.NotNull;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 import lombok.*;
 import org.hibernate.annotations.JdbcTypeCode;
 import org.springframework.data.annotation.CreatedDate;
@@ -26,6 +27,8 @@ public class AmlCheck {
   private Long id;
 
   @ValidPersonalCode private String personalCode;
+
+  private UUID companyId;
 
   @Enumerated(value = EnumType.STRING)
   @NotNull
--- a/src/main/java/ee/tuleva/onboarding/aml/AmlKybCheckEventListener.java
+++ b/src/main/java/ee/tuleva/onboarding/aml/AmlKybCheckEventListener.java
@@ -2,11 +2,15 @@ package ee.tuleva.onboarding.aml;
 
 import static java.util.Map.entry;
 
+import ee.tuleva.onboarding.company.Company;
+import ee.tuleva.onboarding.company.CompanyRepository;
+import ee.tuleva.onboarding.kyb.CompanyDto;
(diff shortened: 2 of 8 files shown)


stenehrlich-tuleva:
### Addressed the listener-ordering P1 from the Codex review (commit `ec2f2ca02`)

**Problem:** `CompanyOnboardingEventListener` and `AmlKybCheckEventListener` both handle `KybCheckPerformedEvent` with no ordering guarantee, and the `company` row is created **only** by the onboarding listener reacting to that same event. On a company's first passing screening, whenever the AML listener ran first it resolved `company_id = null` and persisted every KYB check unattributed — exactly the path M7's company-id-keyed risk view depends on.

**Fix:**
- New `KybCheckPerformedEventOrder` constants (mirrors the existing `SebEventListenerOrder` pattern): `ONBOARD_COMPANY` (100) runs before `ATTRIBUTE_AML_CHECKS` (200).
- `@Order` on both listeners, so the company row always exists before AML resolves its id.
- `AmlKybCheckCompanyAttributionIntegrationTest`: publishes a first-time passing screening and asserts the persisted checks carry the newly created `company_id`.

**Verified** under the pg/Testcontainers profile (Docker): passes with the fix; **fails as `company_id = [null, null]`** when `@Order` is removed — confirming the test reproduces the bug and guards against it. The genuine null path (failed screen / not-yet-onboarded company) is preserved as `NO_DATA`.

PR #1707, opened 2026-06-16
Avoid concurrent insertions of party

vootelerotov:
The transactions endpoint and the savings-fund statement endpoint both lazily create the ledger party for a first-time investor via LedgerPartyService.getOrCreate. When a dashboard load fires both concurrently for a not-yet-created party, both insert the same (party_type, owner_id) and one fails with:

  duplicate key value violates unique constraint "ux_party_type_owner"

The previous guard (pg_advisory_xact_lock + double-checked re-read) did not prevent this. An advisory xact lock is held only until its transaction ends, so it serializes callers only while they hold it across their INSERT. getTransactions is @Transactional and does; but SavingsFundStatementService.getAccountStatement is not, so there each statement is its own transaction and the lock is released the instant its SELECT returns -- before the re-check and INSERT. That path holds no lock across the write, so it never excluded the transactional path.

Replace the lock with an idempotent upsert that does not depend on holding a lock across the write: INSERT ... ON CONFLICT DO NOTHING, then re-fetch. This is correct regardless of transaction vs autocommit, connection sharing, or isolation level. Apply the same pattern to the account path (ux_account_owner_name), which would otherwise surface the same race once the party stopped failing.

Verified on both engines: H2 @DataJpaTest idempotency tests (default ./gradlew test) and a 16-thread LedgerPartyConcurrencyIntegrationTest on real PostgreSQL (gated by assumeTrue(isPostgres()), SPRING_PROFILES_ACTIVE=pg,test).

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/ledger/LedgerAccountService.java
+++ b/src/main/java/ee/tuleva/onboarding/ledger/LedgerAccountService.java
@@ -8,7 +8,9 @@ import ee.tuleva.onboarding.ledger.LedgerAccount.AccountType;
 import ee.tuleva.onboarding.ledger.LedgerAccount.AssetType;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 import lombok.RequiredArgsConstructor;
+import org.springframework.jdbc.core.simple.JdbcClient;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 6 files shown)


ErkoRisthein:
Thanks!!

PR #1730, opened 2026-06-25
Make KYC flow role-aware for represented children (TKF PR0)

stenehrlich-tuleva:
PR0 of the TKF child-onboarding feature — makes the KYC/AML flow **role-aware**.

## Problem
KYC keyed off the authenticated person's `userId`/`personalCode` (the parent). When a parent acted as a represented child, the survey, AML screening and `savings_fund_onboarding` all landed on the **parent**, so the child's onboarding never completed and sanctions screening screened the parent. `RedemptionRequest` had the same parent-not-child bug.

## Change — target the active role's party (the child)
- `KycSurveyService.submit` / `getIdentity` resolve the subject from the active role code → write/read under the child
- `KycCheckService` runs the risk check + `BeforeKycCheckedEvent` + `KycCheckPerformedEvent` against the child
- `KycIdentityController` returns the child's identity
- `RedemptionVerificationService` screens the redemption **party** (child), not the actor (kept `userId` = actor for audit)

PEP stays derived from the parent's own KYC (unchanged).

## Gate test
`ChildRoleAwareKycIntegrationTest`: parent switches into the child role → submits KYC → the **child's** `savings_fund_onboarding` flips to `COMPLETED`, and the AML screening + identity read target the child, not the parent. Risk is driven via `TestKycChecker`, so role-targeting is proven independently of the DB risk rule.

> End-to-end child `COMPLETED` against the live DB also needs the `database-administration` change removing "minor without 3rd pillar = medium" from `kyc_ob_assess_user_risk` (plan decision 20).

Plan: `TulevaEE/TKF-VPII2026` → `plaanid/lapse-tkf-klient-rahvastikuregister-plaan.md` §0.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyService.java
+++ b/src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyService.java
@@ -3,6 +3,7 @@ package ee.tuleva.onboarding.kyc.survey;
 import ee.tuleva.onboarding.auth.principal.AuthenticatedPerson;
 import ee.tuleva.onboarding.country.Country;
 import ee.tuleva.onboarding.kyc.KycCheckService;
+import ee.tuleva.onboarding.user.User;
 import ee.tuleva.onboarding.user.UserService;
 import java.util.Optional;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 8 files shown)


src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyService.java
vootelerotov, line 44: These two methods now work differently, one resolves subject, other does not, but they have exactly the same name. Lets do `getRoleIdentity`?

vootelerotov (approved):
One naming idea.

src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyService.java
vootelerotov: Resolved by just inlining the secnod method -- we do not call it anywhere. Also removes the need to resolve USER from database, we already have the user here (that we resolve).

PR #1744, opened 2026-06-30
Allow OPS to manually create guardian (eestkoste) links for adult wards

stenehrlich-tuleva:
## What

Adds a guardian (eestkostja → eestkostetav) representation path to the existing `POST /admin/parent-child-link` ops endpoint, so OPS can manually link a guardian to a limited-capacity ward — **including adults aged 18+**, which the current parent/minor path rejects.

Planning / context: TulevaEE/TKF-VPII2026#99

## Why

TKF "lapsele kogumine" only models parent↔minor representation today (`RepresentationType.LEGAL_REPRESENTATIVE`, ward must be < 18, `validUntil` = 18th birthday). A guardian of a person with limited legal capacity — often an **adult** under court-appointed eestkoste — couldn't be set up at all. Most adults have full capacity, but these exceptions need to be handled manually by OPS.

## How

- `RepresentationType.GUARDIAN` — distinguishes guardianship from parental representation (auditable in data).
- `ParentChildLinkRegistrationService.registerGuardian(...)` — no minor/age check; takes an explicit `validUntil` from OPS (required, must be strictly in the future). OPS is instructed to set a far-future date.
- `CreateParentChildLinkRequest` — new `validUntil` field + `@AssertTrue` requiring it for `GUARDIAN`.
- `AdminController` branches on `relationshipType`; the parent/minor path is unchanged.
- `InvalidGuardianshipValidUntilException` → HTTP 400.
- **No DB migration**: `relationship_type` is `text` and `valid_until` stays `NOT NULL`. Role resolution (`RoleSwitchService`) and payment verification (`PaymentVerificationService`) work unchanged via the existing `validUntil > today` gate — so a far-future date keeps the guardian link active, unlike the parent link that expires at age 18.

## OPS process (not in code)

OPS verifies the court order (eestkoste + varahooldusõigus), then posts the guardian as `parentCode`, the ward as `childCode`, `relationshipType: GUARDIAN`, and a far-future `validUntil`.

## Tests

- Service: adult ward creates a `GUARDIAN` link; past and today `validUntil` rejected; parent path still rejects 18+.
- Controller: guardian request delegates to `registerGuardian` + seeds onboarding; missing `validUntil` → 400.

## Out of scope

- Self-service UI (stays OPS-manual for these rare exceptions).
- Open-ended (`NULL`) `validUntil` — OPS sets an explicit far-future date for now.
- Renaming `parent`/`child` fields to `representative`/`represented` (separate refactor).

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/admin/AdminController.java
+++ b/src/main/java/ee/tuleva/onboarding/admin/AdminController.java
@@ -371,14 +371,38 @@ public class AdminController {
         request.parentCode(),
         request.childCode(),
         request.childFirstName(),
-        request.childLastName(),
-        request.relationshipType());
+        request.childLastName());
     savingsFundOnboardingService.seedPersonOnboardingIfAbsent(request.childCode());
 
     return "Created parent-child link: parentCode="
         + request.parentCode()
         + ", childCode="
         + request.childCode();
+  }
+
+  @PostMapping("/guardian-link")
+  public String createGuardianLink(
+      @RequestHeader("X-Admin-Token") String token,
+      @Valid @RequestBody CreateGuardianLinkRequest request) {
+
+    validateTokenWithOpsAccess(token);
+    if (!request.validUntil().isAfter(LocalDate.now(clock))) {
+      throw new ResponseStatusException(
+          BAD_REQUEST,
+          "Guardian link validUntil must be in the future: validUntil=" + request.validUntil());
+    }
+    parentChildLinkRegistrationService.registerGuardian(
+        request.guardianCode(),
+        request.wardCode(),
+        request.wardFirstName(),
+        request.wardLastName(),
+        request.validUntil());
+    savingsFundOnboardingService.seedPersonOnboardingIfAbsent(request.wardCode());
+
+    return "Created guardian link: guardianCode="
+        + request.guardianCode()
+        + ", wardCode="
+        + request.wardCode();
   }
 
   @PostMapping("/blackrock-adjustment")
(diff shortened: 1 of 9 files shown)


src/main/java/ee/tuleva/onboarding/admin/AdminController.java
```diff
@@ -353,12 +354,21 @@ public String createParentChildLink(
       @Valid @RequestBody CreateParentChildLinkRequest request) {
 
     validateTokenWithOpsAccess(token);
-    parentChildLinkRegistrationService.register(
-        request.parentCode(),
-        request.childCode(),
-        request.childFirstName(),
-        request.childLastName(),
-        request.relationshipType());
+    if (request.relationshipType() == GUARDIAN) {
```
vootelerotov: I am not a fan of this, splitting the logic here. What about adding a separate `@PostMapping("/guardian-link")`?

vootelerotov: Can make valid until mandatory then, do not need to make it conditional.

src/main/java/ee/tuleva/onboarding/party/ParentChildLinkRegistrationService.java
```diff
@@ -36,31 +37,56 @@ public ParentChildLink register(
...
+  public ParentChildLink registerGuardian(
+      String guardianPersonalCode,
+      String wardPersonalCode,
+      String wardFirstName,
+      String wardLastName,
+      LocalDate validUntil) {
+
+    if (!validUntil.isAfter(LocalDate.now(clock))) {
+      throw new InvalidGuardianshipValidUntilException(wardPersonalCode, validUntil);
+    }
```
vootelerotov: I think we should validate on controller. Fundamentally, it does not seem like it is problematic that we have historical guardianship recorded. But in ops controller we should flag this, as it is quite likely a human error?

vootelerotov: If tihs would come from another source than manual entry, I would think we would flag this?

vootelerotov (approved):
I think that forcing guardian to be a parent might be too narrow.

But we can also accept this limitation for now, and manually handle any edge cases.

PR #1749, opened 2026-07-02
Don't reject onboarded companies when beneficial-owner data is missing

ErkoRisthein:
On 2026-07-01 the business registry stopped returning beneficial-owner (tegelik kasusaaja) rows in the ettevottegaSeotudIsikud_v1 response, so every ownership gate check failed and the nightly KYB monitoring flipped all 139 COMPLETED legal entities to REJECTED. The savings account statement endpoint then swallowed the resulting "Not onboarded" exception and returned 204 No Content, so companies saw 0 balances, and one company payment was returned as "not onboarded".

- LegalEntityOnboardingEventListener no longer downgrades a COMPLETED company when the only failing gate checks are ownership checks with no evidence of an actual change; it logs an error for manual review instead. Real changes (owner personal code, ownership percentage, structure swap) still reject, and new companies still cannot complete onboarding without beneficial-owner verification.
- KybDataChangeDetector change entries now carry a metadataChanged flag and KybCheckPerformedEvent exposes hasOwnershipEvidenceChange(), so the change evidence is computed before the event is published instead of re-reading persisted check history in a listener (which would race with the AML listener that persists the current run).
- AccountStatementController no longer swallows exceptions as 204: SavingsFundStatementService returns Optional, empty means 204 (no savings account), and errors propagate as 500 with a stack trace.
- SavingsFundStatementService returns the real balance for any party with a ledger FUND_UNITS account regardless of onboarding status, via the new read-only LedgerService.findPartyAccount (no account-creation side effects for parties that never had one). Pay-ins and pay-outs for rejected companies remain blocked by the existing verification checks.

- After each nightly KYB monitoring run, a single digest message goes to the AML Slack channel (KybMonitoringCompletedEvent -> LegalEntityManualReviewNotifier): the count and registry codes of COMPLETED companies whose ownership checks failed in that run and were kept for manual review. No message when there are none.

Known gap: DUAL/SINGLE_BOARD ownership check metadata does not record owner identities, so a same-total owner swap while beneficial-owner data is unavailable is flagged for manual review rather than auto-rejected. Do not close it by enriching stored metadata shapes - the next nightly comparison would flag every company as changed and mass-downgrade again; add a self-baselining identity fingerprint check instead.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/account/SavingsFundStatementService.java
+++ b/src/main/java/ee/tuleva/onboarding/account/SavingsFundStatementService.java
@@ -47,18 +47,22 @@ public class SavingsFundStatementService {
             fundUnitsAccount ->
                 statement(
                     account ->
-                        ledgerService
-                            .findPartyAccount(ownerCode, partyType, account)
-                            .map(LedgerAccount::getBalance)
-                            .orElse(BigDecimal.ZERO)));
+                        account == FUND_UNITS
+                            ? fundUnitsAccount.getBalance()
+                            : ledgerService
+                                .findPartyAccount(ownerCode, partyType, account)
+                                .map(LedgerAccount::getBalance)
+                                .orElse(BigDecimal.ZERO)));
   }
 
   private FundBalance statement(Function<UserAccount, BigDecimal> balances) {
+    BigDecimal nav = getNAV();
+
     BigDecimal units = balances.apply(FUND_UNITS).negate();
-    BigDecimal value = getNAV().multiply(units).setScale(2, HALF_UP);
+    BigDecimal value = nav.multiply(units).setScale(2, HALF_UP);
 
     BigDecimal reservedUnits = balances.apply(FUND_UNITS_RESERVED).negate();
-    BigDecimal reservedValue = getNAV().multiply(reservedUnits).setScale(2, HALF_UP);
+    BigDecimal reservedValue = nav.multiply(reservedUnits).setScale(2, HALF_UP);
 
     return FundBalance.builder()
         .fund(getSavingsFund())
--- a/src/main/java/ee/tuleva/onboarding/account/SavingsFundStatementService.java
+++ b/src/main/java/ee/tuleva/onboarding/account/SavingsFundStatementService.java
@@ -8,12 +8,16 @@ import static java.math.RoundingMode.HALF_UP;
 import ee.tuleva.onboarding.auth.principal.AuthenticatedPerson;
 import ee.tuleva.onboarding.fund.Fund;
 import ee.tuleva.onboarding.fund.FundRepository;
+import ee.tuleva.onboarding.ledger.LedgerAccount;
 import ee.tuleva.onboarding.ledger.LedgerParty.PartyType;
 import ee.tuleva.onboarding.ledger.LedgerService;
+import ee.tuleva.onboarding.ledger.UserAccount;
 import ee.tuleva.onboarding.savings.fund.SavingsFundConfiguration;
 import ee.tuleva.onboarding.savings.fund.SavingsFundOnboardingService;
 import ee.tuleva.onboarding.savings.fund.nav.FundNavProvider;
 import java.math.BigDecimal;
+import java.util.Optional;
+import java.util.function.Function;
 import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Service;
 
@@ -27,18 +31,33 @@ public class SavingsFundStatementService {
   private final FundRepository fundRepository;
   private final SavingsFundConfiguration savingsFundConfiguration;
 
-  public FundBalance getAccountStatement(AuthenticatedPerson person) {
+  public Optional<FundBalance> getAccountStatement(AuthenticatedPerson person) {
     String ownerCode = person.getRoleCode();
     PartyType partyType = PartyType.from(person.getRoleType());
 
-    if (!savingsFundOnboardingService.isOnboardingCompleted(person.toPartyId())) {
-      throw new IllegalStateException("Not onboarded: code=" + ownerCode);
+    if (savingsFundOnboardingService.isOnboardingCompleted(person.toPartyId())) {
+      return Optional.of(
+          statement(
+              account ->
+                  ledgerService.getPartyAccount(ownerCode, partyType, account).getBalance()));
     }
+    return ledgerService
+        .findPartyAccount(ownerCode, partyType, FUND_UNITS)
+        .map(
+            fundUnitsAccount ->
+                statement(
+                    account ->
+                        ledgerService
+                            .findPartyAccount(ownerCode, partyType, account)
+                            .map(LedgerAccount::getBalance)
+                            .orElse(BigDecimal.ZERO)));
+  }
 
-    BigDecimal units = getUnits(ownerCode, partyType);
+  private FundBalance statement(Function<UserAccount, BigDecimal> balances) {
+    BigDecimal units = balances.apply(FUND_UNITS).negate();
     BigDecimal value = getNAV().multiply(units).setScale(2, HALF_UP);
 
-    BigDecimal reservedUnits = getReservedUnits(ownerCode, partyType);
+    BigDecimal reservedUnits = balances.apply(FUND_UNITS_RESERVED).negate();
     BigDecimal reservedValue = getNAV().multiply(reservedUnits).setScale(2, HALF_UP);
 
     return FundBalance.builder()
@@ -48,8 +67,8 @@ public class SavingsFundStatementService {
         .value(value)
         .unavailableUnits(reservedUnits)
         .unavailableValue(reservedValue)
-        .contributions(getSubscriptions(ownerCode, partyType))
-        .subtractions(getRedemptions(ownerCode, partyType))
+        .contributions(balances.apply(SUBSCRIPTIONS).negate())
+        .subtractions(balances.apply(REDEMPTIONS).negate())
         .build();
   }
 
@@ -60,25 +79,6 @@ public class SavingsFundStatementService {
           "Savings fund not found: isin=" + savingsFundConfiguration.getIsin());
     }
     return fund;
-  }
-
-  private BigDecimal getUnits(String ownerCode, PartyType partyType) {
-    return ledgerService.getPartyAccount(ownerCode, partyType, FUND_UNITS).getBalance().negate();
-  }
-
-  private BigDecimal getReservedUnits(String ownerCode, PartyType partyType) {
-    return ledgerService
-        .getPartyAccount(ownerCode, partyType, FUND_UNITS_RESERVED)
-        .getBalance()
-        .negate();
-  }
-
-  private BigDecimal getSubscriptions(String ownerCode, PartyType partyType) {
-    return ledgerService.getPartyAccount(ownerCode, partyType, SUBSCRIPTIONS).getBalance().negate();
-  }
-
-  private BigDecimal getRedemptions(String ownerCode, PartyType partyType) {
-    return ledgerService.getPartyAccount(ownerCode, partyType, REDEMPTIONS).getBalance().negate();
   }
 
   private BigDecimal getNAV() {
--- a/src/main/java/ee/tuleva/onboarding/kyb/KybDataChangeDetector.java
+++ b/src/main/java/ee/tuleva/onboarding/kyb/KybDataChangeDetector.java
@@ -43,13 +43,18 @@ public class KybDataChangeDetector {
     for (var current : currentChecks) {
       var previous = previousByType.get(current.type());
       if (isExistingCheck(previous) && changed(previous, current)) {
-        changes.add(change(current.type(), previous.success(), current.success()));
+        changes.add(
+            change(
+                current.type(),
+                previous.success(),
+                current.success(),
+                metadataChanged(previous, current)));
       }
     }
 
     for (var previous : previousChecks) {
       if (isRemovedCheck(previous, currentByType)) {
-        changes.add(change(previous.type(), previous.success(), "N/A"));
+        changes.add(change(previous.type(), previous.success(), "N/A", true));
       }
     }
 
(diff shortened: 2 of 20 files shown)


src/main/java/ee/tuleva/onboarding/kyb/KybDataChangeDetector.java
```diff
@@ -43,13 +43,18 @@ public KybCheck detect(
...
+                previous.success(),
+                current.success(),
+                metadataChanged(previous, current)));
       }
     }
 
     for (var previous : previousChecks) {
       if (isRemovedCheck(previous, currentByType)) {
-        changes.add(change(previous.type(), previous.success(), "N/A"));
+        changes.add(change(previous.type(), previous.success(), "N/A", true));
```
ErkoRisthein: Not a fan of boolean arguments like that. Any better alternative that complies with Clean Code?

src/test/groovy/ee/tuleva/onboarding/kyb/KybScreeningIntegrationTest.java
```diff
@@ -229,6 +234,44 @@ void rescreeningAnUnchangedSoleOwnerCompanyDoesNotFlagDataChanged() {
     assertThat((List<Map<String, Object>>) dataChanged.metadata().get("changes")).isEmpty();
   }
 
+  @Test
+  void missingBeneficialOwnerEvidenceKeepsCompletedCompanyCompleted() {
+    var registryCode = "12345678";
+    var owner = boardMemberOwner(PERSONAL_CODE, 100.0).build();
+    kybScreeningService.screen(companyWith(owner));
+    assertThat(onboardingRepository.findStatus(registryCode, PartyId.Type.LEGAL_ENTITY))
+        .contains(SavingsFundOnboardingStatus.COMPLETED);
```
ErkoRisthein: Use static imports here for better readability. `LEGAL_ENTITY`, `COMPLETED`

ErkoRisthein (commented):
Automated high-effort review (8 finder angles → dedup → adversarially verified; 36 raw candidates, 21 survived, top 10 posted inline).

**Blocking for tonight's deploy: the 3 P0/P1 comments** — the `Instant` bind kills the AML digest on prod PostgreSQL, the stale removed-check evidence re-rejects 2 of the 139 companies on the next nightly run, and the unguarded notifier fails the whole job on any alerting hiccup.

Not posted (verified but below the cap / accepted trade-offs): fail-then-pass digest false positive within one run window, true-zero 200s for former account holders (matches the intended semantics), the deliberate dual alert stream (Sentry + AML digest), and ownershipPercent source instability (affected 0 of 139 in last night's data).

src/main/java/ee/tuleva/onboarding/savings/fund/SavingsFundOnboardingRepository.java
```diff
@@ -29,6 +31,32 @@ public Optional<SavingsFundOnboardingStatus> findStatus(String code, PartyId.Typ
...
+                WHERE c.registry_code = o.code
+                  AND ac.type IN ('KYB_SOLE_MEMBER_OWNERSHIP',
+                                  'KYB_DUAL_MEMBER_OWNERSHIP',
+                                  'KYB_SINGLE_BOARD_MEMBER_OWNERSHIP')
+                  AND ac.success = false
+                  AND ac.created_time >= :since
+              )
+            ORDER BY o.code
+            """)
+        .param("since", since)
```
ErkoRisthein: **[P0 · CONFIRMED — blocks merge] Raw `Instant` bind fails on prod PostgreSQL.**

Spring JDBC 7's `StatementCreatorUtils` has no `Instant` mapping and pgjdbc 42.7 rejects `setObject(Instant)` with *"Can't infer the SQL type to use for an instance of java.time.Instant"*. H2 accepts it, so tests stay green while the nightly AML digest query crashes in production (and the exception propagates into the monitoring job). Every other `JdbcClient` timestamp bind in the codebase wraps with `Timestamp.from(...)` (`NavLedgerRepository`, `TkfVolumeReader`).

Fix: `.param("since", Timestamp.from(since))`.

src/main/java/ee/tuleva/onboarding/kyb/KybDataChangeDetector.java
ErkoRisthein: **[P1 · CONFIRMED — re-rejects 2 real companies tonight] Stale removed-check entries count as fresh ownership-change evidence.**

`AmlKybCheckHistory.getLatestChecks` returns *all* rows from the trailing year, and this loop emits `metadataChanged=true` for every historical row whose type is absent from the current run. A company that restructured SOLE→DUAL months ago re-emits a stale `SOLE_MEMBER_OWNERSHIP: N/A` entry nightly; during the current RIK outage its failing DUAL check (unchanged metadata) then looks like "evidence of change" and the listener auto-REJECTs instead of keeping for review. Last night's prod `KYB_DATA_CHANGED` metadata shows 2 of the 139 companies with exactly these `N/A` entries — they get re-rejected on the next run even after the revert.

Suggested fix: in the listener, only treat change entries for the *currently failing* check types as evidence (a stale removed-type entry then no longer poisons the decision).

src/main/java/ee/tuleva/onboarding/savings/fund/LegalEntityManualReviewNotifier.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Component
+@RequiredArgsConstructor
+class LegalEntityManualReviewNotifier {
+
+  private final SavingsFundOnboardingRepository savingsFundOnboardingRepository;
+  private final OperationsNotificationService notificationService;
+
+  @EventListener
+  public void onKybMonitoringCompleted(KybMonitoringCompletedEvent event) {
```
ErkoRisthein, line 24: **[P1 · CONFIRMED] Unguarded synchronous listener — a digest failure fails the whole monitoring job.**

If the query throws (see the `Instant` bind bug) or Slack is down, the exception propagates out of `publishEvent` in `KybMonitoringService.screenAllCompanies` and the scheduled job is marked failed *after* all screening succeeded. Alerting should never fail the job it reports on — wrap the body in try/catch + `log.error` (same principle as the TD-never-blocks-NAV rule).

src/main/java/ee/tuleva/onboarding/savings/fund/SavingsFundOnboardingRepository.java
```diff
@@ -29,6 +31,32 @@ public Optional<SavingsFundOnboardingStatus> findStatus(String code, PartyId.Typ
...
   }
 
+  public List<String> findCompletedLegalEntitiesWithFailedOwnershipChecksSince(Instant since) {
+    return jdbcClient
+        .sql(
+            """
+            SELECT o.code
+            FROM savings_fund_onboarding o
+            WHERE o.type = 'LEGAL_ENTITY'
+              AND o.status = 'COMPLETED'
```
ErkoRisthein, line 44: **[P2 · CONFIRMED] `status = 'COMPLETED'` filter makes wrongly-REJECTED companies invisible.**

Any company flipped to REJECTED during an ongoing outage permanently drops out of this digest (and the listener early-returns on REJECTED→REJECTED), so there is no in-code recovery or visibility path — only the manual revert SQL. Operationally fine for the current incident *if* the revert runs before tonight's screening; worth a follow-up to also surface still-REJECTED legal entities with recent ownership failures.

src/main/java/ee/tuleva/onboarding/savings/fund/LegalEntityOnboardingEventListener.java
```diff
@@ -52,10 +62,17 @@ public void onKybCheckPerformed(KybCheckPerformedEvent event) {
...
-        .allMatch(KybCheck::success);
+        .filter(check -> check.type().isOnboardingGate() && !check.success())
+        .toList();
+  }
+
+  private boolean isInconclusiveOwnershipFailure(
+      List<KybCheck> failedGateChecks, KybCheckPerformedEvent event) {
+    return !failedGateChecks.isEmpty()
+        && failedGateChecks.stream().allMatch(check -> check.type().isOwnershipCheck())
+        && !event.hasOwnershipEvidenceChange();
```
ErkoRisthein: **[P2 · CONFIRMED — accepted gap, documenting] History reset by a new representative masks a genuine owner change.**

Check history is keyed by `(personalCode, companyId)`. When the company is sold and the new owner submits a fresh KYB survey, `getLatestChecks` returns empty → `DATA_CHANGED` has no entries → a failing ownership gate looks inconclusive → kept COMPLETED. Mitigated by the manual-review flag + AML digest; the durable fix (identity comparison independent of submitter) belongs with the tegelikudKasusaajad REST API migration on July 10.

src/main/java/ee/tuleva/onboarding/savings/fund/LegalEntityOnboardingEventListener.java
```diff
@@ -28,12 +28,22 @@ public void onKybCheckPerformed(KybCheckPerformedEvent event) {
...
+    if (oldStatus == COMPLETED && isInconclusiveOwnershipFailure(failedGateChecks, event)) {
+      log.error(
+          "KYB ownership verification inconclusive, keeping completed status for manual review: registryCode={}, personalCode={}, failedChecks={}",
+          registryCode,
+          personalCode,
+          formatFailedChecks(event.getChecks()));
+      return;
+    }
+
     savingsFundOnboardingRepository.saveOnboardingStatus(registryCode, LEGAL_ENTITY, newStatus);
```
ErkoRisthein, line 47: **[P2 · CONFIRMED] COMPLETED→REJECTED downgrade only logs INFO.**

When the listener *does* auto-reject a previously-completed company (evidence changed / non-ownership gate failure), the only trace is `log.info` — no Sentry/Slack. Unnoticed rejection of a live customer is exactly what made this incident bad; a one-line `log.error` when `oldStatus == COMPLETED && newStatus == REJECTED` gives it the same visibility as the keep-path.

src/main/java/ee/tuleva/onboarding/savings/fund/LegalEntityOnboardingEventListener.java
```diff
@@ -52,10 +62,17 @@ public void onKybCheckPerformed(KybCheckPerformedEvent event) {
...
-        .filter(check -> check.type().isOnboardingGate())
-        .allMatch(KybCheck::success);
+        .filter(check -> check.type().isOnboardingGate() && !check.success())
+        .toList();
+  }
+
+  private boolean isInconclusiveOwnershipFailure(
+      List<KybCheck> failedGateChecks, KybCheckPerformedEvent event) {
+    return !failedGateChecks.isEmpty()
+        && failedGateChecks.stream().allMatch(check -> check.type().isOwnershipCheck())
```
ErkoRisthein, line 80: **[P3 · CONFIRMED — future-proofing note] Guard keys on WHICH check failed, not on registry-data completeness.**

If RIK's next degradation drops shareholder/board rows instead of beneficial-owner rows, ownership screeners emit no checks and `COMPANY_STRUCTURE` fails instead — a non-ownership failure that bypasses this guard and mass-rejects again. The durable fix is an explicit registry-data-completeness signal (e.g. from the new REST API's `mustSubmitBeneficialOwners`/`hiddenBeneficiariesNumber` fields) rather than enumerating check types. Fine to defer to the July 10 migration.

src/main/java/ee/tuleva/onboarding/account/SavingsFundStatementService.java
```diff
@@ -27,18 +31,33 @@ public class SavingsFundStatementService {
...
-    if (!savingsFundOnboardingService.isOnboardingCompleted(person.toPartyId())) {
-      throw new IllegalStateException("Not onboarded: code=" + ownerCode);
+    if (savingsFundOnboardingService.isOnboardingCompleted(person.toPartyId())) {
+      return Optional.of(
+          statement(
+              account ->
+                  ledgerService.getPartyAccount(ownerCode, partyType, account).getBalance()));
     }
+    return ledgerService
+        .findPartyAccount(ownerCode, partyType, FUND_UNITS)
```
ErkoRisthein, line 45: **[P3 · CONFIRMED] Redundant I/O in the statement path.**

`getNAV()` is called twice per statement (two identical latest-NAV lookups; a republish between them prices `value` and `unavailableValue` differently in one response), and the non-onboarded path fetches FUND_UNITS twice plus re-resolves the `LedgerParty` for each of the four accounts (~10 ledger queries per GET). One local NAV variable + reusing the already-fetched account/party fixes both. (The unused `fundUnitsAccount` lambda parameter is the smell marker.)

src/main/java/ee/tuleva/onboarding/savings/fund/SavingsFundOnboardingRepository.java
```diff
@@ -29,6 +31,32 @@ public Optional<SavingsFundOnboardingStatus> findStatus(String code, PartyId.Typ
...
+              AND EXISTS (
+                SELECT 1
+                FROM aml_check ac
+                JOIN company c ON ac.company_id = c.id
+                WHERE c.registry_code = o.code
+                  AND ac.type IN ('KYB_SOLE_MEMBER_OWNERSHIP',
```
ErkoRisthein: **[P3 · CONFIRMED] Ownership check types hardcoded in SQL — third copy of this list.**

This IN-list duplicates `KybCheckType.isOwnershipCheck()` (and the aml module's type mapping). If a fourth ownership check type is added, the listener keeps such companies COMPLETED as inconclusive but this query never reports them — they silently vanish from the AML digest. Derive the names from `KybCheckType.values()` and bind as a collection param.

src/main/java/ee/tuleva/onboarding/savings/fund/LegalEntityManualReviewNotifier.java
```diff
@@ -0,0 +1,32 @@
...
+import ee.tuleva.onboarding.kyb.KybMonitoringCompletedEvent;
+import ee.tuleva.onboarding.notification.OperationsNotificationService;
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+class LegalEntityManualReviewNotifier {
```
ErkoRisthein, line 18: **[P3 · CONFIRMED — convention] Missing `@NullMarked`.**

CLAUDE.md: *"New packages should be @NullMarked from the start; when touching an unmarked file, mark its package (or just the class)"* — this new class compiles with no NullAway coverage. Add `@NullMarked` on the class (the repository change would ideally get the same treatment).

PR #1761, opened 2026-07-06
Fix custody verification for parents with both personal and property custody

stenehrlich-tuleva:
## The bug
`CustodyVerificationService.verify` rejected legitimate full-`varahooldus` parents as `NOT_ASSET_MANAGEMENT`, so child onboarding always fell to "under review".

The population register returns **personal (H10)** and **property (H20)** custody as **separate entries** for the same child — a full-custody parent has *both*, and RR lists **personal first**. The code took the first matching entry with `findFirst()` and checked only that one:

```java
CustodyRight custody = fetchCustodyRights(parent).stream()
    .filter(r -> r.childPersonalCode().equals(childCode))
    .findFirst().orElse(null);          // ← grabs H10 (personal)
...
if (!custody.grantsAssetManagement())   // PERSONAL != PROPERTY → false
    return notVerified(NOT_ASSET_MANAGEMENT);
```

So the H20 asset-management right sitting in the *next* list element was never checked. This would reject essentially **every** full-custody parent.

## Confirmed against live RR
A live X-Road `/isikud` probe returned exactly two custody entries for the child — `H10 täielik isikuhooldusõigus` then `H20 täielik varahooldusõigus`, both `staatus=H1` (kehtiv), child `olek=E` (alive). The mapper constants were all correct; the verification logic was the fault.

## The fix
Look for the property (asset-management) right **across all** of the child's custody entries instead of inspecting whichever comes first. `NO_CUSTODY` / `CHILD_NOT_ALIVE` / `NOT_ASSET_MANAGEMENT` precedence is preserved.

## Test
Added `verifiesWhenChildHasBothPersonalAndPropertyCustodyEvenWhenPersonalIsListedFirst` — reproduces the real RR ordering (personal before property, same child). It **fails on the old `findFirst` code** and passes with the fix. The existing "verifies" test never caught this because it listed a *different* child first, so `findFirst` never tripped. Full `CustodyVerificationServiceTest` + spotless green.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/party/CustodyVerificationService.java
+++ b/src/main/java/ee/tuleva/onboarding/party/CustodyVerificationService.java
@@ -8,7 +8,9 @@ import static ee.tuleva.onboarding.party.CustodyVerification.Outcome.OK;
 import ee.tuleva.onboarding.populationregister.CustodyRight;
 import ee.tuleva.onboarding.populationregister.PopulationRegisterClient;
 import ee.tuleva.onboarding.populationregister.PopulationRegisterPerson;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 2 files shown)


vootelerotov (approved):
LGTM!

PR #1780, opened 2026-07-16
Include the child's name in the eligible-children listing

stenehrlich-tuleva:
## What

`GET /v1/me/children` returned only personal codes, so the child-onboarding step-1 dropdown could only offer a parent a list of raw isikukoodid to choose between. This adds the child's first and last name to each entry.

## Why it's cheap

The population register's `hooldusoigused` (custody) block — which the eligible-children lookup already queries — can return the other person's name (`TeineIsikEesnimi` / `TeineIsikPerenimi`) alongside the code. Tuleva's RR access is authorised for those two fields, verified against **production** RR before writing this. So the name adds **zero extra X-Road calls**.

## Changes

- Request `TeineIsikEesnimi`/`TeineIsikPerenimi` in `PersonQueryRequest.DataFields.custody()`.
- Carry the name on `CustodyRight` (nullable — the register doesn't always populate it, and the custody decision never depends on it; a compact 4-arg constructor keeps existing call sites unchanged).
- Surface it through `findChildrenWithAssetManagementCustody` → `findEligibleChildren` → new `EligibleChild` → `EligibleChildResponse`.
- Names arrive uppercase from the register and are capitalised the same way onboarding already stores them (`ParentChildLinkRegistrationService` via `capitalizeFully`).

## Tests

- `PopulationRegisterClientTest` — new case: the two name fields are requested, and `SIIM-JÜRI`/`JÕEORG` map to `Siim-Jüri`/`Jõeorg`.
- `ChildOnboardingServiceTest`, `CustodyVerificationServiceTest`, `ChildControllerTest` updated for the name-carrying signatures.
- 81 tests across `party` + `populationregister` green; spotless clean.

Frontend counterpart (render the name in the dropdown) is a separate PR on onboarding-client.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/party/ChildOnboardingService.java
+++ b/src/main/java/ee/tuleva/onboarding/party/ChildOnboardingService.java
@@ -31,12 +31,15 @@ public class ChildOnboardingService {
   private final ApplicationEventPublisher applicationEventPublisher;
   private final Clock clock;
 
-  public List<String> findEligibleChildren(AuthenticatedPerson parent) {
+  public List<EligibleChild> findEligibleChildren(AuthenticatedPerson parent) {
     LocalDate today = LocalDate.now(clock);
     return custodyVerificationService
         .findChildrenWithAssetManagementCustody(parent.getPersonalCode(), CUSTODY_MAX_AGE)
         .stream()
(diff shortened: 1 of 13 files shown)


vootelerotov (approved):
LGTM!

PR #1786, opened 2026-07-16
Accept multiple investment goals in a KYC survey

stenehrlich-tuleva:
## What

The KYC survey's `INVESTMENT_GOALS` item held a **single** value. This makes it a **list**, so a saver can pick more than one goal (e.g. education *and* a first home).

## Backward compatible

`@JsonFormat(with = ACCEPT_SINGLE_VALUE_AS_ARRAY)` on the list means a single value still deserializes into a one-element list. So:
- Live **person / company** flows keep sending a single value — untouched, no coordinated deploy needed.
- **Already-stored** surveys (single value in jsonb) still read back.
- Only the **child** flow sends a real array (separate onboarding-client PR, stacked on #1595).

Nothing in the KYC code reads the investment goal as a single value (it's stored as jsonb and only ever deserialized), so no consumer breaks.

## Tests

- New: multiple goals in one item deserialize to a 2-element list; a single value still deserializes (backward compat).
- Existing JSON-payload survey tests pass unchanged (the annotation absorbs the single-value shape).
- `KycSurveyResponseTest` 10 green; spotless clean.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyResponseItem.java
+++ b/src/main/java/ee/tuleva/onboarding/kyc/survey/KycSurveyResponseItem.java
@@ -1,5 +1,6 @@
 package ee.tuleva.onboarding.kyc.survey;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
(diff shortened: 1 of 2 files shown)


vootelerotov (approved):
LGTM.

PR #1796, opened 2026-07-20
Second parent (co-guardian) pending access — backend

stenehrlich-tuleva:
Backend slice for letting the **second parent** of a child see and join the child's savings-fund account. Draft — compliance sign-off (Maria) gates the merge, not the review.

## What it does
When one parent opens a child's account, we capture the **other** guardian(s) from the population register and create a `PENDING_KYC` `parent_child_link` for each. That link **grants no access** — it only lets the other parent *discover* the child (`GET /v1/me/pending-child-onboardings`) and later *join* via their own onboarding/KYC, which activates the link.

## The invariant (a pending link must never authorize anything)
Enforced at the **repository** so it can't be forgotten at a call site: the three "active representation" / "other active parents" `@Query` methods bake in `status = ACTIVE`. Verified every consumer inherits it:
- **Role switch** — `RoleSwitchService.getRoles` (ACTIVE-only finder) + `switchToRepresentedChild` guard via `isActiveRepresentation`.
- **Parent payments** — `PaymentVerificationService.isAuthorizedRemitter` → `isActiveRepresentation` (now ACTIVE-only).
- **Notification sender** — `ParentChildLinkNotificationSender.otherActiveParents` reads the repo directly; that finder is now ACTIVE-only.

The unfiltered `…AndRelationshipType` finder is used **only** by registration/activation, never on an auth path. Tests prove each path excludes PENDING (`ParentChildLinkRepositoryTest`, `RoleSwitchServiceTest#switchToPendingChildIsRejected`, `builderDefaultsStatusToActive`).

## How the co-parent is found
A child-subject custody query — `fetchCustodyRights(requester, subject)` — returns both guardians (RR already returns the full record; verified live). The cache is **bypassed** for it so each requester's X-Road call is distinct and audited. A direction-neutral `Guardian` type is used instead of overloading the parent-oriented `CustodyRight` (whose `childPersonalCode`/`childAlive` fields are wrong-way-round for a child subject).

## Deviations flagged for review
1. **Activation is on `KycCheckPerformedEvent`, not `SavingsFundOnboardingCompletedEvent`.** The child's aggregate completion fires once (first parent only); the KYC-check event fires per parent, so activation stays per-co-parent and endpoint-agnostic. It gates on `PERSONAL_ONBOARDING` + risk `LOW/NONE` (MEDIUM/HIGH never activate). The acting parent isn't on the event (subject is the child), so it's read from the synchronous request's `SecurityContext`.
2. **⚠️ Known end-to-end gap (the piece to design next):** the existing KYC survey is submitted by a parent *role-switched to the child* — but the invariant (correctly) blocks role-switching to a PENDING child, so a co-parent **cannot yet drive the child's KYC** through the current endpoint, and their activation won't fire. All backend building blocks are in place (`resolvePendingChildCode` — scoped join resolution; `activate`; the completion listener). The missing piece is the **frontend join flow + a dedicated join endpoint** that mints a scoped child capability from the resolved pending id, so the co-parent can complete the child's KYC without a prior active role-switch. Deliberately did **not** activate on onboarding-start (that would grant access before KYC).
3. **`registerPending` uses portable find-first + save, not `INSERT … ON CONFLICT DO NOTHING`** (H2 test DB lacks `ON CONFLICT`). Same guarantees (idempotent, never downgrades ACTIVE, silent). Residual: a same-instant race where two guardians open the *same* child could hit the unique constraint; `ChildOnboardingService` swallows it (log + continue) so the opening parent's onboarding is unaffected.

## Tests
`compileJava`/`compileTestJava` clean; targeted suites green (`ParentChildLink*`, `Custody*`, `RoleSwitch*`, `ChildOnboarding*`, `PopulationRegister*` — 122 tests, incl. a new `ParentChildLinkActivationListenerTest`). One transient H2 flake (shared in-memory schema drop across many Spring contexts, a known repo infra issue) did not reproduce on re-run.

## Next
Frontend PR (switcher merge + join flow) + the join endpoint that closes gap #2.


stenehrlich-tuleva:
Reviewed with **codex** (diff review) + the **Tuleva pre-PR checklist**, and addressed everything in \`f1548056d\`:

**codex**
- **P1** — a captured co-parent who went through the normal custody-verified onboarding stayed stuck `PENDING_KYC` (couldn't switch/pay), because `findOrCreateLink` returned the existing row unchanged. Fixed: `findOrCreateLink` now **activates** a pending link on re-registration (flip → ACTIVE + publish event), so the co-parent becomes an active representative exactly like the first parent. This also collapses the earlier "gap #2" — the co-parent just re-runs the existing flow, no bespoke capability endpoint needed — so the now-redundant `KycCheckPerformedEvent` activation listener (and its SecurityContext read) was **removed**.
- **P2** — the best-effort `registerPending` try/catch didn't actually isolate failures (a `@Transactional` failure marks the outer tx rollback-only). Fixed: co-parent capture moved to `ChildOnboardedEvent` → `ChildCoParentCaptureListener` with `@TransactionalEventListener(AFTER_COMMIT)`, so it runs only after the opening parent's onboarding commits and can never roll it back (nor orphan pending links).

**Tuleva checklist**
- `spotlessApply` (was failing); trimmed explanatory comments off implementation classes per AGENTS.md (kept two one-liners: the ACTIVE-status invariant on the repository, the cache-bypass reason on the RR client).
- Clean: branch (not master), thin controller, migration `text` + explicit named constraint + `V1_230` above max, events published after save, no IAE→500, `populationregister` `@NullMarked`. Minor/deferred: new files in the unmarked `party`/`auth.role` packages aren't NullAway-checked (marking them is a separate refactor).

Gates green in-hand: `spotlessCheck` ✓, `compileJava`/`compileTestJava` ✓, targeted tests ✓ (incl. `ChildCoParentCaptureListenerTest` proving a capture failure is swallowed).

src/main/java/ee/tuleva/onboarding/auth/role/RoleController.java
```diff
@@ -19,6 +19,12 @@ public List<Role> getRoles(@AuthenticationPrincipal AuthenticatedPerson person)
     return roleSwitchService.getRoles(person);
   }
 
+  @GetMapping("/v1/me/pending-child-onboardings")
```
ErkoRisthein: What about making it more generic? i.e. pending-onboardings? Since this could be useful for companies as well?

src/main/java/ee/tuleva/onboarding/party/ParentChildLink.java
```diff
@@ -41,6 +41,12 @@ public class ParentChildLink {
   @Column(nullable = false)
   private LocalDate validUntil;
 
+  @NotNull
+  @Builder.Default
+  @Enumerated(STRING)
+  @Column(nullable = false)
+  private ParentChildLinkStatus status = ParentChildLinkStatus.ACTIVE;
```
ErkoRisthein: Use static imports. I.e. `private ParentChildLinkStatus status = ACTIVE;` for less verbosity.

src/main/java/ee/tuleva/onboarding/party/ParentChildLink.java
```diff
@@ -56,4 +62,16 @@ void prePersist() {
...
+  public boolean isPending() {
+    return status == ParentChildLinkStatus.PENDING_KYC;
+  }
+
+  public boolean isActive() {
+    return status == ParentChildLinkStatus.ACTIVE;
+  }
+
+  void activate() {
+    this.status = ParentChildLinkStatus.ACTIVE;
```
ErkoRisthein: Avoid using `this` keyword unless strictly needed, for less verbosity.
`status = ACTIVE;`

src/main/java/ee/tuleva/onboarding/party/ParentChildLinkRepository.java
```diff
@@ -5,18 +5,44 @@
 import java.util.Optional;
 import java.util.UUID;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
 
 public interface ParentChildLinkRepository extends JpaRepository<ParentChildLink, UUID> {
 
+  // Security invariant: the "active representation" queries require status = ACTIVE, so a
+  // PENDING_KYC link never authorizes anything. Do not remove the status filter from these
+  // @Queries.
```
ErkoRisthein: Remove comment

src/main/java/ee/tuleva/onboarding/populationregister/Guardian.java
```diff
@@ -0,0 +1,8 @@
+package ee.tuleva.onboarding.populationregister;
+
+public record Guardian(String personalCode, boolean assetManagement, boolean valid, boolean alive) {
```
ErkoRisthein: Avoid using boolean flags like this since it creates a lot of ambiguity and error-prone situations like:
``` 
new Guardian("12345678901", true, true, true);
new Guardian("12345678901", true, false, false);
new Guardian("12345678901", true, true, false);
``` 
etc.
 Any better way? Enums?

src/main/java/ee/tuleva/onboarding/populationregister/RestPopulationRegisterClient.java
```diff
@@ -78,6 +78,18 @@ public PopulationRegisterResult<List<CustodyRight>> fetchCustodyRights(
         PersonMapper::toCustodyRights);
   }
 
+  @Override
+  public PopulationRegisterResult<List<Guardian>> fetchCustodyRights(
+      String requesterPersonalCode, String subjectPersonalCode, Duration maxAge) {
+    // Never cache child-subject custody: the store keys by subject only, so reuse would serve one
+    // parent's response to another and skip their per-requester X-Road audit. Always fetch fresh.
```
ErkoRisthein, line 85: Why? Can you explain

src/main/resources/db/migration/V1_230__parent_child_link_status.sql
```diff
@@ -0,0 +1,6 @@
+-- ACTIVE authorizes representation; PENDING_KYC grants no access. Existing rows are all real links.
+ALTER TABLE parent_child_link
+    ADD COLUMN status text NOT NULL DEFAULT 'ACTIVE';
+
+ALTER TABLE parent_child_link
+    ADD CONSTRAINT parent_child_link_status_check CHECK (status IN ('ACTIVE', 'PENDING_KYC'));
```
ErkoRisthein: Maybe use enums instead?

src/main/java/ee/tuleva/onboarding/populationregister/RestPopulationRegisterClient.java
ErkoRisthein: The response store is keyed by (subject personal code, query type) only — there is no requester column. For this query the subject is the child but the requester varies: each parent or guardian asks about the same child. With caching enabled, parent B's lookup could be served from parent A's stored response, so B would receive population-register data about the child without ever making their own X-Road call.

That matters because the audit trail lives in the X-Road message log: every fresh call carries `X-Road-UserId = requester`, which is the record of who queried the registry about whom. The other two query types don't have this problem — there the requester is the subject, so caching per subject is caching per requester.

Making this cacheable safely would need a requester column in the response store (and requester-scoped lookups); until someone needs that, always fetching fresh is the simple correct option.

src/main/java/ee/tuleva/onboarding/auth/role/RoleController.java
ErkoRisthein: Done in e12110c74 — the endpoint is now `GET /v1/me/pending-onboardings` returning `PendingOnboardingResponse(type, code, name)`, mirroring the `Role` shape. Companies later just become `LEGAL_ENTITY` entries in the same list, no new endpoint or client contract change needed.

src/main/java/ee/tuleva/onboarding/party/ParentChildLink.java
ErkoRisthein: Done in e12110c74, and da6a0a4f6 sweeps the rest of the branch: static imports for MINOR_CUSTODY_VERIFICATION, ROLE_SWITCH, the RoleType values, ArgumentMatchers.any and APPLICATION_JSON in the touched files. The one place left qualified is `PartyId.Type.PERSON` in RoleSwitchService(+Test) — it clashes with the statically imported `RoleType.PERSON` in the same files.

src/main/java/ee/tuleva/onboarding/party/ParentChildLink.java
ErkoRisthein: Done in e12110c74 — `activate()` is now just `status = ACTIVE;`.

src/main/java/ee/tuleva/onboarding/party/ParentChildLinkRepository.java
ErkoRisthein: Done in e12110c74 — went one step further than deleting the comment: the three `@Query` overrides are now plain derived queries that take `ParentChildLinkStatus` as an explicit parameter, so every caller states `ACTIVE` or `PENDING_KYC` at the call site and there is no hidden filter left to warn about.

src/main/java/ee/tuleva/onboarding/populationregister/Guardian.java
ErkoRisthein: Done — `Guardian` is now `(personalCode, Type custodyType, CustodyValidity custodyValidity, Status personStatus)`, reusing `CustodyRight.Type` and `PopulationRegisterPerson.Status` plus a small `CustodyValidity` enum (e12110c74). Test fixtures read `new Guardian(CO_PARENT, PROPERTY_CUSTODY, VALID, ALIVE)` now. da6a0a4f6 also renames `Type.PROPERTY` → `PROPERTY_CUSTODY` / `PERSONAL` → `PERSONAL_CUSTODY` since a bare statically-imported PROPERTY read like real estate — note new custody-verification evidence records will carry the new spelling.

src/main/resources/db/migration/V1_230__parent_child_link_status.sql
ErkoRisthein: Done in da6a0a4f6 — `parent_child_link_status` is now a proper `CREATE TYPE ... AS ENUM` like the ledger types, mapped with `@JdbcType(PostgreSQLEnumJdbcType.class)`, and the CHECK constraint is gone entirely. (Interim context: the first fix commit rewrote the CHECK as `= ANY(ARRAY[...])` because on H2 2.4 a CHECK with a literal IN list breaks database-wide once the migrating session closes — that made the full local test suite fail on this branch while CI's PostgreSQL stayed green. The enum type sidesteps all of that.) Since V1_230 changed in place, reset your local dev database if you already ran the earlier version.

ErkoRisthein (approved):

PR #1797, opened 2026-07-21
Add manual legal entity onboarding override for KYB edge cases

stenehrlich-tuleva:
## Why

Some legitimate companies can't pass automated KYB — e.g. **Kojanarri OÜ** (12934765): one shareholder (osanik) but two beneficial owners (tegelikud kasusaajad), because one is a beneficial owner by virtue of being a spouse. Per @vootelerotov, we don't want to encode every such edge case in the rules — a manual override is the better tool.

This restores a manual-onboarding capability (the old `/admin/whitelist-company` was removed in `4503056e0`), but as **per-check overrides with a mandatory reason** rather than a blanket status flip — because a bare status flip skips the `company_party` data that onboarding populates, and the periodic KYB re-screen (AML monitoring) would revert it.

## What

- **`kyb_check_override`** (migration `V1_230`) — per-check overrides (`registry_code`, `check_type`, `forced_success`, `reason`, `created_by`), unique per `(registry_code, check_type)`.
- **`KybScreeningService`** applies stored overrides during screening, so the forced check passes through the **normal pipeline**: status → `COMPLETED`, `company_party` rows populated, AML sees the pass. Persisted → the nightly re-screen respects them (no revert, no "backfill required" alert).
- **`ManualCompanyOnboardingService`** (`kyb.survey`) records a **non-attesting** KYB survey on the company's behalf — needed because a never-submitted company has no survey and `screenLatest` would otherwise throw — verifies the person is a board member, persists the override(s), then re-screens.
- **`POST /admin/manually-onboard-company`** (ops-token gated). Only ownership checks + `SELF_CERTIFICATION` are forceable (`KybCheckType.isManuallyForceable`); anything else (sanctions, PEP, eligibility) → `400`. A blank reason → `400`.

## OPS usage

For Kojanarri (never submitted, so `SELF_CERTIFICATION` must also be forced — the operator attesting on the company's behalf, audited via the reason):

```
POST /admin/manually-onboard-company
  X-Admin-Token: <ops-token>
  registryCode=12934765
  personalCode=<board member personal code>
  forcedChecks=SINGLE_BOARD_MEMBER_OWNERSHIP,SELF_CERTIFICATION
  reason=<mandatory justification>
```

## Review

Ran a `tuleva-pr-check` self-review and an independent adversarial review, and hardened the first cut:
- **Was:** the endpoint accepted any `KybCheckType` — could permanently, invisibly force-pass sanctions/PEP. **Now:** allowlisted to ownership + self-cert.
- **Was:** an all-true self-cert survey was fabricated, silently passing `SELF_CERTIFICATION` and overwriting the client's real attestation. **Now:** a non-attesting survey; passing self-cert requires an explicit, audited override.
- Blank reason rejected; test uses a real `User` fixture.

## Needs compliance (@Maria) sign-off

- **Forceable-check allowlist membership** — currently ownership + `SELF_CERTIFICATION`. Should `COMPANY_STRUCTURE` / `RELATED_PERSONS_KYC` be forceable too, or is ownership+self-cert the right ceiling?
- **Self-cert on behalf** — is an operator force-passing `SELF_CERTIFICATION` (attesting for the company, with a reason) acceptable, or should the company always self-declare?
- **Override longevity (F3)** — an override currently never expires and isn't re-bound to the facts that justified it: if the ownership/beneficial-owner set later changes to genuinely high-risk parties, the forced check keeps passing on every re-screen. Worth an expiry or a re-review trigger — flagging as a follow-up.

## Notes

- Ops-only failure modes (`NotBoardMemberException`, "no user found") currently surface as HTTP 500 — thrown before any DB write, so no partial state; can map to 4xx if preferred.


vootelerotov (approved):
LGTM!

PR #1813, opened 2026-07-29
Screen the guardian when onboarding a child and before the child's KYC risk assessment

stenehrlich-tuleva:
## Why

The internal AML screening procedure ("Klientide skriinimine ja monitoorimine", rows 22b–25 of the minor section) forbids onboarding a minor whose representative is high-risk, sanctioned, or a citizen of a high-risk country — but the guardian was never screened anywhere in the child onboarding path (`ChildRoleAwareKycIntegrationTest` even asserted "no AML checks are written against the parent"). The preventive gate itself lives in `kyc_ob_assess_user_risk` (database-administration), where `check_24_minor_representative_high_risk` is currently hardcoded to 0; a follow-up DA migration will implement it and read the data this PR starts producing.

## What

- **`ChildOnboardingService`**: after custody verification, the onboarding parent is sanction/PEP-screened via OpenSanctions, and their citizenship (population-register self-lookup, best-effort) is captured into the CUSTODY_RIGHT evidence as `guardianPersonalCode` + `guardianCitizenship` — the data source for the upcoming check_24 high-risk-country branch.
- **`ChildCoParentCaptureListener`**: the captured co-parent is sanction/PEP-screened right after their PENDING_KYC link is registered (skipped with a log line if they have no user account yet — they are re-screened when they onboard themselves).
- **`GuardianKycScreeningListener` (new)**: when a KYC survey is submitted for a minor, all guardians with unexpired links are re-screened synchronously before the risk assessment runs, so check_24 always sees fresh screening data even when the link was created weeks earlier. Screening scope deliberately includes suspended and PENDING_KYC links — suspension does not cleanse the risk association.
- **`ParentChildLinkService.findGuardianCodes`** + repository finder over the existing `idx_parent_child_link_child` index.

Screening failures (OpenSanctions or population register down) never block the child flow — the risk assessment treats missing data as unknown. Whether missing guardian screening should instead fail closed is an open compliance decision for the DA-side check_24 PR.

## Population-register self-lookup

`fetchPerson(parentCode, parentCode)` queries the parent's own record for citizenship. Confirmed OK by Sten 2026-07-29; the code degrades gracefully (citizenship stays unknown) if the register ever rejects the query.

## Merge order

This PR should deploy **before** the database-administration check_24 migration, so guardian screening data is fresh by the time the gate starts enforcing.

## Performance

No new public endpoints. The new `parent_child_link` finder runs only on minor KYC submissions (authenticated, low volume) and is served by the existing child-code index. Adds 1 population-register call per child onboarding and 1–3 OpenSanctions calls per child onboarding / minor KYC submission — same volume class as the existing child screening on those flows. No caching needed, no N+1, no unbounded result sets.

## Known limitation (deliberate, layer-3 scope)

Co-parent capture is async (`AFTER_COMMIT`): if the child's KYC survey is submitted in the same seconds the capture is still running, the co-parent's link may not exist yet and `GuardianKycScreeningListener` screens only the initiating parent. In practice the capture completes while the parent is still filling the survey, the co-parent is screened before they ever gain access (their PENDING_KYC link grants none), and continuous guardian-risk monitoring after COMPLETED is explicitly out of scope for this change. Flagged by Codex review as P2; accepted and documented rather than gating child KYC on an async external call.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/party/ChildCoParentCaptureListener.java
+++ b/src/main/java/ee/tuleva/onboarding/party/ChildCoParentCaptureListener.java
@@ -2,6 +2,9 @@ package ee.tuleva.onboarding.party;
 
 import static org.springframework.transaction.event.TransactionPhase.AFTER_COMMIT;
 
+import ee.tuleva.onboarding.aml.AmlService;
+import ee.tuleva.onboarding.country.Country;
+import ee.tuleva.onboarding.user.UserService;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -16,6 +19,8 @@ public class ChildCoParentCaptureListener {
 
   private final CustodyVerificationService custodyVerificationService;
   private final ParentChildLinkRegistrationService parentChildLinkRegistrationService;
+  private final UserService userService;
+  private final AmlService amlService;
 
   @Async
   @TransactionalEventListener(phase = AFTER_COMMIT)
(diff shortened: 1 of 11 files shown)


vootelerotov:
We pass to OpenSanctions call different parameters, depending on the path we end up taking (main parent, co-parent).

Not ideal, but decided against handling it here -- it is a bigger re-work, and it makes sense to do it in a follow-up.

vootelerotov (approved):
LGTM.

PR #1814, opened 2026-07-29
Fail closed on payout screening + hold high-risk clients' redemptions for AML review

stenehrlich-tuleva:
## Why

Two gaps found in an AML review against the internal procedure ("Klientide skriinimine ja monitoorimine", sisekord 19 / 13):

1. **Payout screening failed open.** `RedemptionVerificationService.runPersonChecks` passed when `addSanctionAndPepCheckIfMissing` returned an empty list. That happens both when OpenSanctions/yente is unavailable (screening error → empty list) **and** when the person has a standing failed sanction/PEP check whose re-screen outcome is unchanged (no new row written → empty list). In both cases the payout went straight to `VERIFIED`. Sisekord 13 p 10.5.6 requires refusing the transaction when sanctions status cannot be verified.

2. **High-risk clients' payouts were never held for approval.** The procedure requires AML specialist confirmation of payouts for high overall-risk clients ("väljamakse kinnitamine AML spetsialisti poolt"), but `RISK_LEVEL`/`TKF_RISK_LEVEL` were not consulted anywhere in the redemption path — and there was no code path for `IN_REVIEW → VERIFIED` at all, so releasing a held payout meant a direct DB update with no record of who approved it.

## What

- **`AmlService.isSanctionAndPepClear(person, country)`** — runs the screening, fails closed on screening error, then requires the *latest persisted* `SANCTION` and `POLITICALLY_EXPOSED_PERSON_AUTO` checks to be successful (absent record = not clear). Existing overrides keep working: they already flip the persisted check's `success` at write time.
- **`RiskLevelService.isHighRisk(personalCode)`** — latest of (`RISK_LEVEL`, `RISK_LEVEL_OVERRIDE`) or (`TKF_RISK_LEVEL`, `TKF_RISK_LEVEL_OVERRIDE`) with metadata `level = 1`. An override row without a `level` counts as a specialist confirmation (same interpretation the Metabase dashboard uses) and does not hold the payout.
- **`RedemptionVerificationService`** — person redemptions now go to `IN_REVIEW` when screening is not clear **or** the party is high-risk.
- **`POST /admin/redemptions/{id}/approve-review?approvedBy=&reason=`** (ops token) — the missing `IN_REVIEW → VERIFIED` path. When a redemption goes to `IN_REVIEW`, a message is posted to the AML Slack channel with the redemption id and amount (no personal data in Slack), so the specialist has an active signal to start the review. Records `reviewed_by`, `review_reason`, `reviewed_at` on `redemption_request` (migration `V1_237`) before verifying, so the approval is auditable.

## Notes for review

- Legal-entity redemptions are unchanged: company risk lives only in `analytics.v_company_risk_metadata` (never written to `aml_check`), so a company high-risk hold needs the company-risk plumbing first — follow-up.
- `RedemptionReviewService` throws `IllegalArgumentException`/`IllegalStateException` like the neighbouring `RedemptionStatusService`; the admin endpoint validates blank params with `ResponseStatusException` like the other admin endpoints.
- The `approvedBy` param is free text because admin auth is a shared token — same limitation as the Metabase actions. Personalized admin identity is a separate problem.
- ⚡ Performance: verification runs per redemption request (single-digit daily volume); the new lookups are 4 point queries on `aml_check` served by `aml_check_personal_code_index`. The admin endpoint is manual. No new hot paths, no caching needed.

## Tests

- `RedemptionVerificationServiceTest` — screening-not-clear → `IN_REVIEW`, high-risk → `IN_REVIEW`, clear+not-high → `VERIFIED`
- `AmlServiceTest` — fail-closed on screening exception; latest-check semantics (incl. standing failed check with unchanged outcome)
- `RiskLevelServiceTest` — high snapshot, high string-level override, non-high, override-without-level
- `RedemptionReviewServiceTest` + `AdminControllerTest` — approve path, blank reason 400, invalid token 401
- `RedemptionIntegrationTest`, `RedemptionVerificationJobTest` green; `spotlessCheck` green

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
+++ b/src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
@@ -13,6 +13,7 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -30,8 +31,39 @@ public class RiskLevelService {
   private final AmlCheckRepository amlCheckRepository;
   private final ApplicationEventPublisher eventPublisher;
 
+  private static final int HIGH_RISK_LEVEL = 1;
+
   public void runRiskLevelCheck(double mediumRiskIndividualSelectionProbability) {
     scorePillar("III pillar", amlRiskReader, RISK_LEVEL, mediumRiskIndividualSelectionProbability);
+  }
+
+  public boolean isHighRisk(String personalCode) {
+    return latestLevelIsHigh(personalCode, List.of(RISK_LEVEL, RISK_LEVEL_OVERRIDE))
+        || latestLevelIsHigh(personalCode, List.of(TKF_RISK_LEVEL, TKF_RISK_LEVEL_OVERRIDE));
+  }
+
+  private boolean latestLevelIsHigh(String personalCode, List<AmlCheckType> types) {
+    return amlCheckRepository
+        .findFirstByPersonalCodeAndTypeInOrderByCreatedTimeDesc(personalCode, types)
+        .map(check -> Objects.equals(riskLevel(check), HIGH_RISK_LEVEL))
+        .orElse(false);
+  }
+
+  // An override row without a level is a specialist confirmation without a score change and does
+  // not keep the client high-risk.
+  private Integer riskLevel(AmlCheck check) {
+    Object level = check.getMetadata().get("level");
+    if (level instanceof Number number) {
+      return number.intValue();
+    }
+    if (level instanceof String string) {
+      try {
+        return Integer.parseInt(string);
+      } catch (NumberFormatException e) {
+        return null;
+      }
+    }
+    return null;
   }
 
   public void runTkfRiskLevelCheck(double mediumRiskIndividualSelectionProbability) {
--- a/src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
+++ b/src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
@@ -13,7 +13,6 @@ import java.time.Instant;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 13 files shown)


src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
```diff
@@ -30,10 +31,41 @@ public class RiskLevelService {
...
+    Object level = check.getMetadata().get("level");
+    if (level instanceof Number number) {
+      return number.intValue();
+    }
+    if (level instanceof String string) {
+      try {
+        return Integer.parseInt(string);
+      } catch (NumberFormatException e) {
+        return null;
+      }
```
vootelerotov, line 82: Is this realistic? It feels weird to handle errors like this in risk check, it feels safer to err on caution than to say that there is no risk level here, if this should really be something that is ever hit? Fall back to very high risk level and then figure out what is going on here?

vootelerotov (commented):
One iffy spot, if we should ever have malformed risk level, then I would rather not say there is no risk.

If there are non-numerical values that can be used for level that we are aware of, I would rather whitelist them.

src/main/java/ee/tuleva/onboarding/aml/risklevel/RiskLevelService.java
stenehrlich-tuleva: Agreed — fixed in 3a3ccd739. A present-but-unparseable level (and a missing level on an automatic RISK_LEVEL/TKF_RISK_LEVEL snapshot, which should never happen) now falls back to high risk and logs an error, so malformed data holds the payout instead of clearing it.

The one case that still reads as "not high" is a *missing* level on an override row — that's the documented Metabase semantics: the "Rakendatud hoolsusmeetmeid" dashboard action copies metadata without a level field, meaning the specialist reviewed and confirmed without changing the score. Whitelisting beyond that: we're not aware of any legitimate non-numeric values, so everything unparseable now errs high rather than being enumerated.

stenehrlich-tuleva:
@vootelerotov Fixed in 3a3ccd739 — malformed risk-level metadata now errs on the side of caution: a present-but-unparseable `level` (or a missing `level` on an automatic snapshot) falls back to **high risk**, holds the payout, and logs an error so we notice the data problem.

On whitelisting: we're not aware of any legitimate non-numeric values, so instead of enumerating them everything unparseable now reads as high. The single deliberate exception is a *missing* `level` on an override row — the "Rakendatud hoolsusmeetmeid" Metabase action creates exactly that shape and it means "specialist reviewed, score unchanged", so it must not re-hold the client. Covered by tests either way.

vootelerotov (approved):

PR #1816, opened 2026-07-31
Screen against every country a person is tied to

stenehrlich-tuleva:
## Why

Sanctions/PEP screening sent OpenSanctions a **single** country, so a dual citizen was only ever scored against one of their citizenships.

For adults it was worse than that: the one country came from the KYC survey's **address** (`KycSurveyService.extractCountry`). The declared citizenship list was stored and read by the V17 risk rules, but **never reached screening at all** — an Estonian resident holding a second citizenship was screened as purely Estonian. For children it was the register's single `PohiKodakondsus`.

The country is a **query property** for `algorithm=logic-v1`, not a filter (the only hard filters in that URL are the `topics`), so it feeds the match score. Sending every country a person is tied to therefore *widens* what screening can catch — this is a recall improvement, not a narrowing.

## What

- `PepAndSanctionCheckService.match`, `addSanctionAndPepCheckIfMissing` and `isSanctionAndPepClear` take a `Set<Country>`; `OpenSanctionsService` sends them all alongside the standing `"ee"`, lower-cased and sorted so the outbound body is deterministic.
- **Adults**: residence **and** every declared citizenship, via `BeforeKycCheckedEvent`. The integration test shows it — a survey declaring EE+FI now screens against both.
- **Children**: the identity query asks the register for the `kodakondsused` group (one field, `Riik`, ISO numeric-3 like `PohiKodakondsus`), so every citizenship in a valid status is captured and screened.
- **Redemption**: screens on the survey countries **plus** the citizenships the custody check recorded. A child's citizenships come from the register rather than their survey, so without this a minor was re-screened on residence alone — and since the fail-closed gate reads the *latest* rows, that re-screen could overwrite an earlier hit with a pass.
- **Evidence**: `CUSTODY_RIGHT` metadata gains a `citizenships` array **beside** the scalar `citizenship` that database-administration's V23 minor gate reads, so those rules keep working untouched.
- New `Countries` helper drops absent and blank codes instead of building empty `Country` objects, replacing the `new Country(null)` sentinel used by the co-parent, guardian and savings-fund batch paths.

## Operational note — the next backfill run is a re-run for everyone

`ChildAmlBackfillService` now treats a record as complete only when the **plural** key is present, which only this change writes. Every child already processed by #1815 is therefore incomplete again, and the next `POST /admin/child-aml-backfill` re-issues **one population-register fetch and one OpenSanctions match per child**. That is the point — it is how existing children gain their second citizenship — but the operator should expect the volume against two rate-limited external APIs. Register queries are billed (~€0.30/record). Runs are operator-triggered, never scheduled.

## Follow-up, not in this PR

The `citizenships` array is written but **nothing reads it yet**: V23's minor gate still reads the scalar, so a child whose *second* citizenship is high-risk is now screened at OpenSanctions but still passes the risk rule. That needs a database-administration change to be worth anything — worth tracking rather than leaving implicit.

`AmlService` reads the `citizenship`/`citizenships` keys from party's custody metadata by name. Modulith stops the constants being shared, which is a fair signal that `party` should expose a child's citizenships directly; left as-is to keep this PR to one idea.

## Data scope

`kodakondsused` is a new field group from the population register. It sits inside composition 1.1 of access decision IT-31679 (isiku detailandmed, which enumerates `kodakondsus`), the same composition `PohiKodakondsus` already comes from — no new data category, and no extra round-trip: it is one more field group on an existing call.

## Tests

CI green. New coverage: a dual citizen reaching OpenSanctions with every country, casing/dedup, the adult survey path, `recordedCitizenships` including the failed-check and no-check paths, register mapping of multiple citizenships, and the backfill re-verifying children recorded before citizenships were captured.

Reviewed pre-merge by codex and the tuleva-pr-check checklist; both rounds' findings are in the follow-up commits — the sharpest being a `String.valueOf(null)` that would have sent the literal country code `"null"` for any minor whose most recent custody check had failed.


stenehrlich-tuleva:
Nudge on this one @vootelerotov, it has been open 3 weeks.

Context for why it matters: master's interface is still `MatchResponse match(Person person, Country country)` (singular), so for adults the screened country comes from the KYC *address* and a declared second citizenship never reaches screening at all. Nothing merged since 2026-07-31 covers it (#1814 was fail-closed payout screening, #1815 the child backfill).

One thing worth deciding before merge: it forces a full child re-backfill, roughly one register fetch (~0,30 EUR) plus one OpenSanctions match per child. Happy to split that into a follow-up if you would rather land the interface change alone.

vootelerotov:
@stenehrlich-tuleva , sorry, missed this one.

Added couple of commits. Main one being that we would use the same set of countries when running the AML checks all the time. Have a look if you want, merging once you are happy.

vootelerotov (approved):
LGTM!

PR #1817, opened 2026-07-31
Expose the bank account behind each savings fund transaction

stenehrlich-tuleva:
## Why

How a savings fund holding is taxed depends on which bank account the money came from. Units bought from an **investeerimiskonto** are taxed on that account's cash flow (declared as payments in/out); everything else is taxed on the gain at redemption. `/v1/transactions` did not say which account was involved, so the client had no way to tell the two apart and could only ever show the ordinary-route calculation.

## What

`Transaction` gains a `counterpartyIban`:

- **Subscription** → the payment's `remitterIban` (where the money came from)
- **Redemption** → the redemption request's `customerIban` (where it was paid out)

Both resolved with one batched lookup per account type, not per transaction. `SavingFundPaymentRepository.findRemitterIbansByIds` is new; redemptions use the existing `CrudRepository.findAllById`.

Transactions with no external reference (and pension transactions, which build `Transaction` from a `CashFlow`) get `null`.

## Notes

- **Frontend depends on this** — see onboarding-client `tkf-statement-ui`. **Merge this first.**
- Tests caught a real NPE on the way in: `Map.of().get(null)` throws, so any ledger transaction without an external reference would have blown up. Guarded and covered.
- `LedgerAccountFixture.EntryFixture` gained an optional `externalReference`; the existing 2- and 3-arg constructors are unchanged, so no existing test moved.
- ⚡ **Performance:** two extra queries per `getTransactions` call, both keyed on primary keys, no N+1. No new indexes needed. This is an authenticated per-user endpoint, not a public one.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
+++ b/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
@@ -20,7 +20,8 @@ public record Transaction(
     String isin,
     CashFlow.Type type,
     BigDecimal units,
-    BigDecimal nav)
+    BigDecimal nav,
+    String counterpartyIban)
     implements Comparable<Transaction> {
 
   public static Transaction from(CashFlow cashFlow) {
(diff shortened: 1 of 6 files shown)


stenehrlich-tuleva:
Closing: the consumer is gone. oc#1612 moved the tax calculation server-side (statement.ts removed, the page now reads `GET /v1/savings-fund/tax-report`), so nothing is left to read `counterpartyIban`. If the investeerimiskonto vs ordinary-route split is still wanted, it should be cut inside the tax report instead.

PR #1818, opened 2026-07-31
Probe: fetch the population register's service description from X-Road

stenehrlich-tuleva:
## Why

We are still guessing at the register's `andmevaljad` field names:

- `kodakondsused` (plural) — needed to close the dual-citizen gap for children (see #1816, which fixed the same gap for adults)
- `isikuAadressid`, `isikuKontaktid`, `dokumendid` and birth place — the fields Alar asked us to capture

SMIT states that **the OpenAPI is the only specification** ("täiendavaid spetsifikatsioone ei tule, OpenAPI dokumentatsiooni täiendatakse jooksvalt"), and that it lives in the X-tee catalogue — reachable only through a security server. So the field names can't be looked up from a developer machine or the public web.

## What

A gated diagnostic, modelled on the earlier RR probe (#1795), that asks X-Road's **meta-services** instead of asking the register about a person:

- `listMethods` — which service codes the `rr` subsystem actually publishes
- `getOpenAPI?serviceCode=isikud.v1` — the full service description, including every field name and permitted value

It prints the lines matching the field names we're unsure about and writes both responses to `build/rr-service-description/` for grepping.

**No personal data**: meta-services describe the service, not a person. Nothing is queried about anyone, and no personal code is needed to run it.

**Bonus**: `listMethods` settles the open production-config question — `application.yml` points at `domesticDataExchange/v1` while access decision IT-31679 names `isikud.v1`. Worth knowing before the child AML backfill (#1815) makes its first production RR calls.

## How to run

```bash
RR_PROBE_ENABLED=true \
RR_PROBE_SECURITY_SERVER=https://xroad-consumer.tuleva.ee \
POPULATION_REGISTER_CLIENT_ID=EE/COM/14118923/tuleva-fund-management \
./gradlew test --tests '*RrServiceDescriptionProbe' -i
```

Overridable: `RR_PROBE_PROVIDER` (default `EE/GOV/70008440/rr`), `RR_PROBE_SERVICE_CODE` (default `isikud.v1`).

Disabled by default via `@EnabledIfEnvironmentVariable`, so CI stays green — verified SKIPPED locally. Throwaway diagnostic: close it once we have the field list.

### DIFF
--- a/src/test/groovy/ee/tuleva/onboarding/populationregister/RrServiceDescriptionProbe.java
+++ b/src/test/groovy/ee/tuleva/onboarding/populationregister/RrServiceDescriptionProbe.java
@@ -34,14 +34,15 @@ import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
  * ./gradlew test --tests '*RrServiceDescriptionProbe' -i
  * </pre>
  *
- * <p>listMethods also settles which service code actually exists — the production URL currently
- * points at {@code domesticDataExchange/v1} while the access decision names {@code isikud.v1}.
+ * <p>listMethods also settles which service code actually exists.
  */
 @EnabledIfEnvironmentVariable(named = "RR_PROBE_ENABLED", matches = "true")
 class RrServiceDescriptionProbe {
 
   private static final String PROVIDER = env("RR_PROBE_PROVIDER", "EE/GOV/70008440/rr");
-  private static final String SERVICE_CODE = env("RR_PROBE_SERVICE_CODE", "isikud.v1");
+  // listMethods (2026-07-31) reports the service code as domesticDataExchange, exposing
+  // POST /v1/isikud — so the production URL, which ends in domesticDataExchange/v1, is correct.
+  private static final String SERVICE_CODE = env("RR_PROBE_SERVICE_CODE", "domesticDataExchange");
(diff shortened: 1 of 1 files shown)


stenehrlich-tuleva:
Closing — a Gradle test in the repo is the wrong shape for this. The probe makes two GETs with one header against X-Road meta-services; a curl does the same with no code, no review burden and nothing to maintain:

```bash
curl -s -H "X-Road-Client: EE/COM/14118923/tuleva-fund-management" \
  "https://xroad-consumer.tuleva.ee/r1/EE/GOV/70008440/rr/getOpenAPI?serviceCode=domesticDataExchange"
```

Recording the two findings so they aren't lost with the branch:

1. **The service code is `domesticDataExchange`, not `isikud.v1`.** `listMethods` reports `domesticDataExchange` exposing `POST /v1/isikud` (plus `foreignDataExchange` and `KodifikaatorInfo` in the same subsystem). `isikud.v1` is the wording in access decision IT-31679, not the X-Road service code.
2. **The production config is therefore correct.** `application.yml` ends in `.../rr/domesticDataExchange/v1` and the client posts `/isikud`, which composes to exactly the registered endpoint. The "fix before live" note in the TKF plan doc does not apply — nothing needs changing before #1815 makes its first register calls.

Also confirmed: the security server is reachable from a developer laptop, no bastion needed.

PR #1820, opened 2026-08-02
Sentry triage 2026 08

ErkoRisthein:


ErkoRisthein:
Superseded by the current Sentry weekly sweep, which carries every change from this PR plus fixes for three issues found while reviewing it.

**Taken from here, as-is**
- `UserService` concurrent-login race. Verified against the production stack trace for the duplicate `personal_code` error on `POST /oauth/token`: the auth path has no application-level transaction, so `SimpleJpaRepository.save` opens its own transaction that genuinely rolls back, and the re-read works.
- `ReturnCalculator` returning no rate instead of throwing. Also verified: the frontend renders `0.0%` (because `null * 100 === 0`), whereas today a single XIRR failure 500s all of `/v1/returns` and the client zeroes every bar. The existing issue is at ~369 events/30d and was muted rather than fixed, so this is a real improvement.
- Logging a missing auth session below error level.
- The distinct `smart.id.timeout` code, and `new.user.flow.signup.error.email.duplicate` in the expected-code list.

**Taken with changes**
- The Smart-ID timeout log included the personal code. That log reaches Sentry, so the parameter was dropped.
- The timeout catch was widened to the `UserActionException` parent as well, so the refused-confirmation and wrong-verification-code siblings stop hitting the catch-all too.
- `ReturnCalculator` logged only `e.getClass().getSimpleName()`. That collapses the benign `NonconvergenceException` together with every `XirrDetails` validation failure (all transactions on the same day, all non-negative, none at all) which are genuine data defects that would then be indistinguishable in Sentry. It now logs the message too.
- `@Nullable` was applied only to the private methods, leaving `ReturnDto.rate` and `Returns.Return.rate` claiming non-null at the boundary that actually gets serialized. Both are now marked.
- The expected-code list is extracted into a small `ExpectedErrorCodes` class with a data-table spec covering every code and both boundaries. This also drops `RejectionLogLevelTest`, whose shared `ListAppender` accumulated events across methods (flagged in review) and which did not cover four of the codes.

**Not taken**
- Returning `Optional.empty()` for a malformed Montonio token. `PaymentController` maps empty to the `/3rd-pillar-payment` redirect, so a forged or tampered token becomes indistinguishable from a legitimately pending payment, with only a warn and no alert. On `POST /v1/payments/notifications` it silently no-ops. The replacement throws `BadCredentialsException`, matching the existing "Token not verified" idiom in the same class.
- `mobile.id.configuration.error` in the expected list. A broken Mobile-ID configuration is our fault, not the user's, and should keep alerting.

**Bugs found while reviewing, now fixed**
- `parseToken` threw `NullPointerException` rather than `ParseException` when the token was null, so the translation was bypassed and it surfaced as a 500. `MontonioNotificationBody` is an unvalidated record, so an empty body on `POST /v1/payments/notifications` reaches it. This affected both implementations.
- `Set.of(...).contains(null)` throws rather than returning false, so a rejection carrying no error code would have failed inside the error handler itself. The `switch` here handled that with `case null -> false`; the set-based version needed an explicit guard.

Note: the replacement branch is not pushed yet. This branch is untouched and this PR can be reopened at any time.

PR #1826, opened 2026-08-04
Move portfolio valuation and cost basis into the backend

stenehrlich-tuleva:
## Why

Voote, on onboarding-client#1611:

> Kas on mõistlik nii palju arvutusloogikat front-endi panna? Mulle on siiamaani tundunud, et idee on olnud panna loogika backendi, ja teha front-end selles osas "õhukesemaks". Backendi reeglina parem testida, saab logida, erindjuhud tulevad meie pool välja.

Agreed. Two calculations were living in the browser: what a person's holdings were worth on a given date, and what redeemed units cost to acquire. Both are numbers we have to stand behind — the second one goes on a tax return — so they belong here.

## What

**`GET /v1/portfolio?from=&to=`** — values every held fund on each day a price was published, grouped into `SAVINGS_FUND`, `SECOND_PILLAR`, `THIRD_PILLAR`. Returns a summary per group (start, end, paid in, withdrawn, gain, gain %) plus the value of each group on each date. The annualised XIRR from `ReturnsService` is folded in, so the client no longer makes a second call to `/v1/returns`.

**`GET /v1/savings-fund/tax-report?year=&method=`** — gain per redemption under `FIFO` or `WEIGHTED_AVERAGE`, defaulting to weighted average.

The logic sits in two plain classes with no Spring in them — `PortfolioValuation` and `CostBasisCalculator` — so the arithmetic is tested directly rather than through a controller.

## Decisions worth knowing

- **A fund held but not yet priced leaves its group `null` for that date, not `0`.** Zero draws a cliff on the day the price history begins, which reads as a jump in someone's money that never happened.
- **The client still decides which days to drop.** Which bands are showing is view state; if the backend dropped dates itself, switching a band off would need a refetch to get them back. The client filters points where a *visible* band is null — filtering, not valuation.
- **XIRR stays per group.** Rates from different sources cannot be summed, so `annualReturnRate` is per group and the client shows it only when one band is selected.
- **No annual rate for the savings fund** — `PersonalReturnProvider` only computes `SECOND_PILLAR` and `THIRD_PILLAR`. It is `null` there rather than a fabricated number.

## Verification

- **27 tests**: 14 over the valuation engine, 6 over cost basis, 3 over the service wiring, 4 over the two controllers
- The valuation and cost-basis tests are ports of the 20 that covered the same arithmetic in the client, including the FIFO €80.00 vs weighted average €58.96 pair on the same disposal, and the case where a second redemption must price off the average rather than leftover FIFO lots
- ⚡ **Performance:** per held fund, one range query for the period plus one point lookup for the price before it started — no N+1, no full history scan. Values are accumulated in a single forward pass over prices and transactions. Both endpoints are authenticated and per-user, not public.

## Depends on

Nothing. The two client PRs now depend on **this** — onboarding-client#1611 and #1612 both call these endpoints, so this merges and deploys first.

#1817 (`counterpartyIban`) is still what investeerimiskonto separation needs, and now that the calculation lives here that separation becomes a backend change too, with no client release.


stenehrlich-tuleva:
## Note: the PIK-transfer finding does not hold

An automated review flagged that `TransactionService.getTransactions` filters EPIS cash flows down to contributions and subtractions, so `TRANSFER_TO_PIK` / `TRANSFER_FROM_PIK` never reach the valuation — and concluded the `SECOND_PILLAR` group is misstated for members with PIK transfers.

Checked it against epis-service. It does not hold.

PIK rows are produced by `cashToTransaction`, not `unitToTransaction` ([`EpisX14TypeToCashFlowStatementConverter`](https://github.com/TulevaEE/epis-service/blob/master/src/main/java/ee/tuleva/epis/account/converter/EpisX14TypeToCashFlowStatementConverter.java)):

```java
private Transaction cashToTransaction(Cash cash) {
  return Transaction.builder()
      .time(getTime(cash))
      .amount(cash.getAmount())
      .type(inferCashType(cash))
      .comment(cash.getComment())
      .build();
}
```

No `isin`, no `units`, no `nav`. `inferCashType` is the only place the two PIK types are produced, and it is reachable only from this cash path — confirmed by `EpisX14TypeToCashFlowStatementConverterSpec`, where the PIK rows carry an amount and a `"PIK number ..."` comment and nothing else.

A row with no ISIN and no unit count cannot move a per-fund unit position. The units leaving the second pillar travel on a separate **unit** row, which carries an ISIN and units and is already included here. Adding the PIK cash rows on top would double-count the outflow, not fix it.

Remaining uncertainty: this is read off the converter and its spec, not off a real EPIS statement for a member who has actually used PIK. If anyone has one to hand, the thing to confirm is that the transfer has a unit leg.

The class of bug is real even though this instance is not — it is the same shape as the `CONTRIBUTION` miss fixed earlier in this branch: the valuation is built on `/v1/transactions`, which is a filtered display feed rather than a complete position feed. A test reconciling the reconstructed position against the EPIS unit balance (`FundBalance.units`) would catch the whole class. Not doing that here; noting it as the obvious follow-up.

stenehrlich-tuleva:
@vootelerotov Aitäh commitide eest — kõik kolm, mille Sten välja tõi, olid päris vead ja kõik minu tehtud. `CONTRIBUTION` oli neist tõsisem kui kirjeldus ütleb: osakud said miinusmärgi **ja** summa läks tulemivalemis väljamakseks, seega topeltviga igal II samba omanikul.

Sinu commitide peale tuli üks asi juurde, mis tasub üle vaadata:

**Maksuaasta käib nüüd väljamakse järgi, mitte kirjendamise ega hinnastamise järgi** (`fdc439eda`). Sa lisasid `priceTime`-i ja portfell kasutab seda, aga maksuaruanne kasutas endiselt `time`-i. Uurisin, kumb maksuõiguslikult õige on — vastus on kumbki mitte:

- **TuMS § 15 lg 3¹** maksustab investeerimisfondi osaku „tagastamisel ... **saadud väljamakse**"
- **TuMS § 12 lg 1** paneb tulu sellesse perioodi, mil see **saadi**

Ehk laekumise põhimõte. `Transaction` sai `settledTime`, mis tuleb `RedemptionRequest.processedAt`-ist; maksuaruanne filtreerib aasta ja kuvab rea kuupäeva selle järgi. Osakute tarbimine käib endiselt kirjendamise järjekorras, sest see otsustab, millised osakud olid käes.

Kaks kohta, kus tahaksin sinu silma:

1. **`processedAt` on see hetk, kui *meie* algatame väljamakse** (`RedemptionBatchJob.markAsRedeemed`), mitte kui pank inimese kontole kannab. Lähim signaal, mis meil on — aga kas „saadud" tähendab TuMS-i mõttes algatamist või laekumist, ei oska mina öelda. Kui meil on kuskil päris laekumise kinnitus, oleks see õigem allikas.
2. **Filter sai ülemise piiri juurde.** Varem tuli see tasuta sellest, et jalutus lõikas kirjendamise kuupäeva pealt. Nüüd, kui aasta tuleb väljamaksest, kirjendamise piir enam aastat ei piira — 20.12 kirjendatud ja 05.01 makstud lunastus oleks muidu jäänud mõlemasse aastasse.

Eraldi kommentaaris eespool ka põhjendus, miks Codexi PIK-leid ei kehti (cash-read ilma ISIN-i ja osakuteta).

vootelerotov (approved):
LGTM, when tests are green, we go.

PR #1837, opened 2026-08-10
Notify the payer when a savings fund payment is returned

ErkoRisthein:
A parent paid for their child, so the payment was addressed to the child while the parent remitted it. The failure notification targeted the child, and PrincipalService blocks under-18s from self-authenticating, so building the email security context threw and rolled back the whole verification transaction. The TO_BE_RETURNED status change and the UNATTRIBUTED_PAYMENT ledger entry were both discarded, the Slack notification had already gone out, and the one minute job retried forever. The ledger stayed short by the payment amount, so bank reconciliation failed every night.

The success path already notifies the payer rather than the recipient. Do the same on the failure path: the payer is the one whose money is being returned, and a payer is never a minor.

Also stop notification failures from rolling back verification. Any listener exception, an EPIS timeout for example, could otherwise discard committed payment state and restart the retry loop. The integration test stubbed only its own collaborators, so three sibling tests were silently throwing inside the same swallowed path while the build stayed green.


vootelerotov:
https://github.com/TulevaEE/onboarding-service/pull/1838/changes/ab9c0598862cd3ee967b4edcfd9f37c4ad6f4f23 -- very similar changeset here, I think this can be closed.

PR #1845, opened 2026-08-12
Stop adding VAT on top of the depot fee rate

taavipertman-glitch:
## Why

The depot fee rates we store (`investment_fee_rate` / `investment_depot_fee_tier`) already include VAT. `DepotFeeCalculator` multiplied the daily accrual by `(1 + vatRate)` anyway, so `daily_amount_gross` counted VAT twice.

The multiplication sat *after* the rate-source branch, so it applied to tier-sourced and fund-specific rates alike — the tier fallback was not exempt.

## What changed

- `DepotFeeCalculator` no longer applies VAT: one daily amount, `dailyAmountGross == dailyAmountNet`, same shape as `ManagementFeeCalculator`.
- `VatRateProvider` (+ its test) deleted — no callers left. It was the only place VAT existed in the codebase.
- `vatRate` removed from the `FeeAccrual` record; `vat_rate` is no longer written.
- The two consumers of the gross amount now read the net one:
  - `FeeAccrualRepository.getAccruedFeesForMonth` (trade-prep liabilities via `TransactionInputService`)
  - `PeriodicTdAttributionService.computeFeeDragPeriod` (depot fee drag → `investment_td_attribution.depot_fee_drag`)

## Impact on existing data

None in practice: the configured `DEPOT` rate is `0.00000000`, so every `daily_amount_gross` is 0 today, and `MANAGEMENT` rows already had gross == net. The bug would only have surfaced once a non-zero depot rate was configured.

## Follow-up (deliberately not in this PR)

`daily_amount_gross` and `vat_rate` columns are kept — external Metabase questions may still select them. Dropping both (and the now write-only gross field) is a separate change once those readers are confirmed.


taavipertman-glitch:
Superseded by #1847, which is now based on `master` and contains this commit plus the NAV-inclusion policy. Both PRs touch `DepotFeeCalculator`, so merging them as one avoids the conflict churn. Closing this one.

PR #1853, opened 2026-08-13
fix(tracking): derive daily fee drag from actual accruals

taavipertman-glitch:
## Problem

The daily TD check charged `annualFeeRate/365` once per NAV event regardless of the calendar gap, and read the `MANAGEMENT` rate only.

Fees accrue per **calendar** day — [`FeeCalculationService`](src/main/java/ee/tuleva/onboarding/investment/fees/FeeCalculationService.java) loops `day.plusDays(1)` from the last accrual to the position report date. So a Monday NAV carries three days of fee while the check offset one.

That gap leaked into `residual` and `navResidual`: roughly 31% of the management fee ((365−252)/365), about 10bp/yr at a 0.34% rate. That noise sits in exactly the number used to judge whether the attribution is complete.

**The day count is the whole live defect.** An earlier draft of this description also claimed the depot fee was never offset and that 100% of it leaked into residual. That is wrong. On the read replica all 695 `investment_fee_accrual` DEPOT rows carry `annual_rate = 0` and `charged_to_fund = false` from inception: **no depot fee has ever reduced any fund's NAV**, so there was nothing for the check to offset and nothing leaking. Reading depot accruals here is correctness for the day Tuleva starts charging one to a fund, not a repair of a live number.

## Fix

Sum the actual `FeeAccrual` rows over `(previousDate, checkDate]` and divide by the fund's NAV. `TrackingInput.annualFeeRate` becomes `accruedFeeFraction`.

Four details the term has to get right to reconcile with the ledger at all:

**Which amount.** `dailyAmountGross` — the one amount an accrual carries after #1847. `recordDailyFees` posts `roundForLedger(dailyAmountGross)` for every fee type and the AUM deduction sums the same column, so there is no other number the NAV could have taken. There is no VAT step anywhere after the rate: the depot tier rates are already VAT-inclusive and fund management is VAT-exempt.

**Which accruals.** Only those the fund was actually charged. `FeeChargedToFundPolicy` decides that per fund, fee type and date; a fee Tuleva bears is still written to `investment_fee_accrual` as the record of the cost, but nothing is posted to the fund's ledger and the NAV never took it. Subtracting it would claim a deduction that never happened, and the difference would land in `navResidual`, which blocks the NAV report. Resolved **per accrual date** rather than once for the window, for the same reason `PeriodicTdAttributionService` does it that way: a policy that changed mid-window would otherwise apply its closing answer to every day before it. Resolution runs only for the fee types actually present in the window — `resolverFor` throws on an unconfigured pair, and the caller loop in `runChecksAsOf` catches only `IncompletePriceDataException`, so resolving every enum value would let one misconfigured fund abort the checks for every remaining fund.

**Which day's NAV.** `fundReturn` is `todayNav/yesterdayNav − 1`, and `NAV_today = NAV_yesterday × (1 + r) − fee`, so the drag is exactly `fee / NAV_yesterday`. The previous NAV date's total is already fetched for the begin-of-day gate, so dividing by the right day costs nothing. The check date's NAV remains a warned fallback for a fund whose previous NAV date carries no positions — returning zero there would recreate the silent drop below. The fallback triggers on `signum() > 0`, matching the begin-of-day gate: a negative NAV would otherwise be accepted as a denominator and turn `feeDrag` positive.

**No silent drops.** `dailyAmountGross` is `NOT NULL`, so the accrual sum no longer filters nulls out of the stream — a row that somehow arrived empty must fail, not be quietly skipped.

**Missing accruals are no longer silent.** An empty window used to yield `feeDrag = 0` with nothing logged, and the whole fee reappeared in `residual` and `navResidual` as if the fund had underperformed. Backfilling a date that predates the accrual table lands there; so does a fee transaction that rolled back while the NAV persisted.

Coverage is measured **per fee type, against the days that fee type was actually charged to the fund** — not against the window length. For each fee type present in the window, the denominator is the count of window days the policy resolver answers `chargedOn`, and the numerator is that type's distinct accrual dates surviving the policy filter.

Both halves of that rule are load-bearing:

- *Per fee type*, because counting distinct dates across all fee types cannot detect what this check was written for. DEPOT rows are written every calendar day, so they alone fill the window and a completely missing `MANAGEMENT` accrual would produce no warning at all. Only a totally empty window warned.
- *Against charged days rather than window days*, because a policy that flips mid-window legitimately charges a fee on a strict subset of the window. Under a plain window-length denominator that correct outcome warns — a brand-new alert that fires on good data. #1851 in this same release exists because a channel of this kind is already being flooded with warnings that misdescribe why a number moved; adding another crying-wolf alert would train people to ignore exactly the signal this check needs to be believed for. Charged days make the flip case complete by construction, and leave the genuinely missing accrual day the only thing that fires.

A window with no accrual rows at all still warns on its own line — that is the backfill / rolled-back-transaction case, and it carries no fee type because no accrual arrived to name one.

**The calculation fails fast; the diagnostic does not.** `FeeChargedToFundPolicy` throws on a gap in the policy rather than guessing an answer, and that stays true everywhere the answer feeds a number: `accruedFeeFraction` resolves the policy for each accrual date and lets an unresolvable one propagate, exactly as `FeeCalculationService` does.

The coverage check is the one place that must not. It now asks the resolver about *every* day of the window, including days no accrual reached, so it can reach a date the calculation never touches. A helper whose entire job is to report a problem must not be able to abort the daily check for the three funds behind it in the loop — a diagnostic that causes an outage is strictly worse than the incomplete coverage it was written to detect. It catches the `IllegalStateException`, logs that coverage is undetermined for that fund and fee type naming the unresolved date, and returns.

Live production has all eight fund × fee-type policy rows open-ended (`valid_to = NULL`), so no gap exists today. The split is here for the same reason as the false-positive fix above: this release is already removing one alarm that cries wolf, and it should not add a path where a warning can take four funds' checks down with it.

Live coverage is in fact complete: `MANAGEMENT` runs unbroken from inception through 2026-08-14 for all four funds, with only the trailing 2026-08-15→08-18 outstanding because the job is behind. So the silent-zero path is real but currently unexercised.

How much a gap would matter: the live `TRACKING_BREACH_THRESHOLD` is **0.001** (10bp, global, effective 26.06.2026 — tightened from 50bp four days earlier). A Friday→Monday gap at a 0.34%/yr rate leaks ≈2.8bp, so a missing window will not trip the gate on its own, but it consumes about a quarter of the breach budget while looking like tracking error. A warning is the right treatment; failing the check would be too much.

This also brings Java in line with the GAS NAV sheet, which already derived fee drag from the `LIABILITY_FEE` balance change (inherently spanning weekends). Java was the outlier, not the sheet.

Reading accruals rather than re-deriving from a rate is also immune to the monthly settlement dip that the GAS script has to special-case.

`navResidual` inherits the correction since it derives from `feeDrag`. `FeeRateRepository` is no longer needed by the service. The fee fraction is computed at the same scale 6 the calculator rounds to, so no dead precision is carried between them.

## Scope

The NAV-time gate needs no separate change — `NavPublisher` → `NavTrackingDifferenceGate.check()` → `trackingDifferenceService.checkFund()` is this same code path.

## Base: #1847

This PR is **based on #1847**, not on master, and already carries its vocabulary — `dailyAmountGross` and `FeeChargedToFundPolicy` — rather than deferring them to whoever merged second. #1847 has to merge and deploy first: it is the change that collapses the accrual to one amount and introduces the policy, and without it this does not compile.

Rebased onto #1847's tip so the diff is only this PR's own commits and the four files they touch. It previously carried duplicates of six master commits, which made it read as a 29-file change.

## Transition note

`feeDrag` is persisted into the event result JSON and summed across the escalation lookback as `escalationFeeDrag`. An escalation window spanning the deploy will mix old `annualRate/365` values with new accrual values. Bounded and harmless: at most the escalation lookback (10 days), only across an unbroken breach streak, and `escalationFeeDrag` is notifier text — it gates nothing.

## Tests

- `TrackingDifferenceCalculatorTest.feeDragIsTheAccruedFractionItselfNotAnAnnualRateSpreadOverADay` — calculator level
- `TrackingDifferenceServiceTest.feeDragUsesActualAccrualsCoveringManagementAndDepot` — a charged depot accrual is included (−0.000012 vs the −0.000009 from management alone)
- `TrackingDifferenceServiceTest.feeDragSpansEveryCalendarDaySincePreviousNavDate` — Friday→Monday picks up all three accrual days
- `TrackingDifferenceServiceTest.feeDragExcludesAFeeTheFundIsNotCharged` — the live depot case: accrued, not charged, not subtracted
- `TrackingDifferenceServiceTest.feeDragFollowsThePolicyInForceOnEachAccrualDate` — a policy that flips mid-window
- `TrackingDifferenceServiceTest.feeDragIsMeasuredAgainstThePreviousNavDatesTotal` — 900k previous vs 1M current, so the two denominators give different answers
- `TrackingDifferenceServiceTest.feeDragIsZeroWhenTheWindowHasNoAccrualsAtAll` — the backfill case
- `TrackingDifferenceServiceTest.aFeeTypeCoveringEveryDayDoesNotMaskAnotherFeeTypesMissingAccrualDay` — DEPOT covers all three days, MANAGEMENT misses one, and the warning still fires for MANAGEMENT only
- `TrackingDifferenceServiceTest.aFeeTypeChargedOnOnlyPartOfTheWindowByPolicyIsNotAnIncompleteAccrual` — the mid-window flip charges DEPOT on one of three days and warns for nothing; paired with the test above, the two together are the specification
- `TrackingDifferenceServiceTest.aFeePolicyGapOnAnAccrualDateFailsTheFeeDragCalculation` — a gap on a day that has an accrual still throws out of the check
- `TrackingDifferenceServiceTest.aFeePolicyGapOnADayWithNoAccrualIsReportedByTheCoverageWarningNotThrown` — the same gap on a day the calculation never touches is logged with the unresolved date and the run completes; paired with the test above, the two together are the split

The tests that pin new behaviour were verified red against the previous behaviour before being kept. Full suite green locally.


taavipertman-glitch:
Rebased onto #1847 and retargeted at `depot-fee-nav-inclusion`, because the two collided in both directions.

**Compile:** this PR read `FeeAccrual::dailyAmountNet`, which #1847 removes along with the `daily_amount_net` column. Whichever merged second would not have compiled.

**Semantics, which mattered more:** #1847 establishes that a fee with `charged_to_fund = false` must not appear in the TD drag, and applies it to `PeriodicTdAttributionService`. The daily fee drag introduced here is a second accrual consumer and did not consult the policy. After both landed it would have subtracted a depot accrual the NAV never bore, and that difference flows into `navResidual` → `NavTrackingDifferenceGate` → a blocked NAV report. Zero today only because every depot rate row is 0 — the `TIER` rows #1847 opens on 18.09.2026 are what turn it into a real number.

The drag now maps `dailyAmountGross` and filters each accrual through `FeeChargedToFundPolicy`, resolved per accrual date rather than once for the window (same reason `PeriodicTdAttributionService` does it that way — a policy that changed mid-window would otherwise apply its closing answer to every day before it).

Two tests added: `feeDragExcludesAFeeTheFundIsNotCharged` and `feeDragFollowsThePolicyInForceOnEachAccrualDate`. `feeDragMatchesTheAmountThatWasPostedToTheLedger` kept its name but lost its VAT premise — there is no net/gross pair left to choose between.

**Base is `depot-fee-nav-inclusion`, so this needs #1847 merged first.** Retarget to `master` after that lands.

PR #1865, opened 2026-08-17
Sum both R17 unit columns when parsing PEVA switching flows

taavipertman-glitch:
# Sum both R17 unit columns when parsing PEVA switching flows

R17 splits each row's units into `Osakud (teenustasuta)` and `Osakud (teenustasuga)`; the row total
is their sum, which the report's own `Hind` and `Summa` columns confirm. The parser read only the
fee-bearing column, so every fee-free unit was dropped.

For the 01.09.2026 cycle that flipped TUK00's switching net from an outflow to an inflow,
understating `liquidityRequired` by 394,520.61 EUR and `paymentLimit` by 365,711.83 EUR in
`PevaRavaFlowService`, which `TransactionInputService` adds to fund liabilities.

Every existing R17 fixture carried identical values in both columns, so the assertions held
whichever column was read — the tests encoded the bug. Fixtures now split their values across the
two columns with the same expected totals.

## Blast radius: nothing persisted (checked against production)

`epis_report_summary` is empty (0 rows), and `investment_report` holds only SEB
`PENDING_TRANSACTIONS`/`POSITIONS` and Swedbank `POSITIONS` — there is no R17 row anywhere in the
database. Flows are computed from the persisted `EpisReportSummary`, not from the raw CSV, so with
no persisted R17 there is no PEVA flow that was computed under the buggy parser: nothing to restate,
and no cycle advanced on a wrong number. The first ingest after this merge is the first one that
counts.

## What else is in the change

**A cross-check, because nothing checked the parser's own output.** R17 carries `Hind` and `Summa`
per row, so units x price must reproduce `Summa`. Had the check existed, the dropped fee-free column
would have failed the very first ingest instead of feeding a wrong liquidity figure into trade
preparation. It throws rather than warns: ingestion is a manual admin upload, so a hard failure is
visible to the operator who can re-export, while a log warning would be missed and the wrong figure
would still be persisted.

**Only our own rows are judged.** Both row checks now run below the `FundResolver` skip. R17 carries
rows for other managers' funds, which we drop anyway; rejecting the whole upload over one would
leave the operator with nothing to fix. This is a behaviour change beyond the cross-check: the
100M-unit sanity limit moved with it, so a *foreign* fund's row above 100M units no longer rejects
the upload. That matters more than it looks now that the number measured is the two columns added
together — a large manager's row can pass 100M when neither of its own columns did. Our own rows
still throw, and the message names the fund.

**`findValue` prefers an exact column name over a longer one that merely contains it.** "summa" is
also inside "Summa (PF valitseja)". The cross-check resolved to the right column only by accident of
column order; a reordered export would have made it read 0.00 and reject every row. The exact pass
is in `EpisCsvParser` and is shared by all the EPIS parsers, so the full suite mattered more here
than the two new tests.

**A zero `Summa` means "not applicable", not a mismatch** — the way `Summa (PF valitseja)` is 0.00
on every row. GAS has always skipped it; Java threw, and Java's was the behaviour that rejects an
upload.

**A row with units and a reported amount but no usable price now fails loudly.** Previously a blank
or non-positive `Hind` silently skipped the cross-check, which is exactly where a dropped units
column could still hide. A report shape that has no `Hind`/`Summa` columns at all (the older
single-column format) still parses — there is nothing to cross-check against there.

## Tolerance: corrected numbers

Earlier descriptions of this branch said the tolerance was 0.5% of the reported amount with a 1 EUR
floor. It is not:

    tolerance = units x 0.000005, floored at 0.02 EUR

i.e. half a price step at 5 decimals, times the units. `Summa` is computed from an unrounded price
while the report shows a rounded one, so the largest legitimate divergence is half a price step per
unit; a real row's cent of rounding still fits under the 0.02 EUR floor.

The claim that "the smallest real divergence was 55% of the row amount" was also wrong, and is the
reason the percentage tolerance had to go. `R17ReportParserTest`
`catchesAMissingUnitComponentTooSmallForAPercentageTolerance` demonstrates **0.29%**: 12,643,992
units x 1.6167 = 20,441,541.87 against a reported 20,501,430.35, a gap of 59,888.48 EUR. A 0.5%
tolerance on that row is 102,507 EUR — it would have passed a divergence of exactly the kind this PR
exists to catch. The new tolerance on the same row is 63.22 EUR.

The price step is a fixed 5 decimals rather than `price.scale()`, because deriving it from the
exporter's formatting is not safe: a price printed as `0.8` instead of `0.80` widens the tolerance
10x, and one that `parseNumber` reads as thousands-grouping (`1,234` under `PERIOD_DECIMAL` → `1234`,
scale 0) widens it to `units x 0.5` — again wide enough to hide an entire dropped column.

## Known limitations, still open

**The real R17 header row is unverified.** The two Estonian column names and the single header row
come from fixtures, not from a genuine EPIS export, and there is no persisted R17 in the database to
check them against. They are deliberately left as they are; guessing at them would be worse than the
bug this PR fixes. Verifying them against a real EPIS file is the one outstanding item.

**The `pfvalitseja` lookup is still resolved by column order.** The exact-match pass fixes the
`summa` collision because a column is literally named `Summa`, but no header equals `pfvalitseja` —
the real one is `PF valitseja/PIK` — so that lookup still falls through to the contains pass and
picks whichever of `PF valitseja/PIK` and `Summa (PF valitseja)` comes first. An export that
reordered them would silently misclassify PIK redemptions as switching flows, and the amount
cross-check would not catch it, because the units on the row are unchanged. Fixing it needs the real
header text, which we do not have.

## Tests

`R17ReportParserTest` covers: summing the two columns, the single-populated-column (older format)
case, the units-vs-amount cross-check, a divergence too small for a percentage tolerance, a
divergence hidden by a price printed with two decimals, a zero amount read as N/A, rounding
tolerated, a missing price on a row that has an amount, foreign-fund rows ignored by both row checks
including above the sanity limit, our own row still throwing above it, and the amount column read
correctly when `Summa (PF valitseja)` comes first. `EpisReportIngestionServiceTest` fixtures were
made internally consistent. Full suite green.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/epis/parser/EpisCsvParser.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/epis/parser/EpisCsvParser.java
@@ -79,9 +79,24 @@ public class EpisCsvParser {
     return group + " " + sub;
   }
 
+  /**
+   * The value of the first column matching any keyword, an exact column name winning over a longer
+   * one that merely contains it.
+   *
+   * <p>Without the exact pass the answer depends on column order, and R17 has two collisions:
+   * "summa" is also inside "Summa (PF valitseja)", and "pf valitseja" is inside it as well. Both
+   * resolve correctly today only because the rows are a LinkedHashMap in column order and the
+   * shorter column happens to come first. An export that reordered them would have made the amount
+   * cross-check read 0.00 and reject every row.
+   */
   public static @Nullable String findValue(Map<String, String> row, String... keywords) {
     for (String keyword : keywords) {
       String normalizedKeyword = normalize(keyword);
+      for (Map.Entry<String, String> entry : row.entrySet()) {
+        if (entry.getKey().equals(normalizedKeyword)) {
+          return entry.getValue();
+        }
+      }
       for (Map.Entry<String, String> entry : row.entrySet()) {
         if (entry.getKey().contains(normalizedKeyword)) {
           return entry.getValue();
--- a/src/main/java/ee/tuleva/onboarding/investment/epis/parser/EpisCsvParser.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/epis/parser/EpisCsvParser.java
@@ -79,28 +79,26 @@ public class EpisCsvParser {
     return group + " " + sub;
   }
 
-  /**
-   * The value of the first column matching any keyword, an exact column name winning over a longer
-   * one that merely contains it.
-   *
-   * <p>Without the exact pass the answer depends on column order, and R17 has two collisions:
-   * "summa" is also inside "Summa (PF valitseja)", and "pf valitseja" is inside it as well. Both
-   * resolve correctly today only because the rows are a LinkedHashMap in column order and the
-   * shorter column happens to come first. An export that reordered them would have made the amount
-   * cross-check read 0.00 and reject every row.
-   */
   public static @Nullable String findValue(Map<String, String> row, String... keywords) {
     for (String keyword : keywords) {
       String normalizedKeyword = normalize(keyword);
-      for (Map.Entry<String, String> entry : row.entrySet()) {
-        if (entry.getKey().equals(normalizedKeyword)) {
-          return entry.getValue();
-        }
+      String exactColumnMatch = row.get(normalizedKeyword);
+      if (exactColumnMatch != null) {
+        return exactColumnMatch;
       }
-      for (Map.Entry<String, String> entry : row.entrySet()) {
-        if (entry.getKey().contains(normalizedKeyword)) {
-          return entry.getValue();
-        }
+      String containsMatch = valueOfColumnContaining(row, normalizedKeyword);
+      if (containsMatch != null) {
+        return containsMatch;
+      }
+    }
+    return null;
+  }
+
+  private static @Nullable String valueOfColumnContaining(
+      Map<String, String> row, String normalizedKeyword) {
+    for (Map.Entry<String, String> entry : row.entrySet()) {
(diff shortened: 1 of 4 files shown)


taavipertman-glitch:
Critical review of the guard added in `bb2049f97` — two things found by reading it rather than running it, both fixed in `288017b01`.

**The tolerance was wide enough to hide the bug it guards against.** 0.5% of the amount is 100,000 EUR on a 20M row, and a missing units column need not be anywhere near that big — a component worth 0.3% of the row is 60,000 EUR and passed cleanly. The allowance now comes from the **price's own precision**: `Summa` is computed from an unrounded price while the report shows a rounded one, so the largest legitimate divergence is half a price step times the units. On that same row it is 632 EUR — 160× tighter — and a real row's cent of rounding still fits.

**A zero amount is not a mismatch.** In R17 a zero `Summa` means "not applicable", the way `Summa (PF valitseja)` is `0.00` on every row. GAS has always skipped it; Java threw. Same guard, two behaviours — and Java's was the one that rejects the whole upload.

Also corrected `doesNotMisinterpretPeriodDecimalUnitsAsThousandsGrouping`, which had 100.500 units at 0.80 against a `Summa` of 80.00. It only passed because the tolerance was loose. The row is consistent at 80.40 now, so the tolerance is not load-bearing for a test about decimal parsing — same principle as reverting the foreign-fund fixture in `00dbdc352`.

Both new tests verified red against the previous parser. Full suite green: 6,592 tests, 0 failures, 61 skipped.

The matching GAS changes are in TulevaEE/tuleva#316, so the two implementations agree on both points.

PR #1866, opened 2026-08-17
Publish the price source priority so SQL consumers cannot disagree with us

taavipertman-glitch:
`PriorityPriceProvider.PRICE_FEEDS` decides which source wins when two carry the **same price date**. The NAV price query in the tuleva repo (`apps/nav-calc/instrumentide-värsked-hinnad.sql`) had its own copy of that ordering, hardcoded as a `CASE` — and the two had already drifted:

| | this service | the NAV query |
|---|---|---|
| 1 | BlackRock | BlackRock |
| 2 | Morningstar | StockMKT (exchange) |
| 3 | **EODHD** | Morningstar |
| 4 | Deutsche Börse | **EODHD** |
| 5 | Euronext | — |
| 6 | Yahoo | *(absent — the NAV must not use Yahoo)* |

For an **ETF** priced on the same date by both the exchange and EODHD, the query picks the exchange and this service picks EODHD — opposite answers. Nothing looks wrong today only because the sources currently agree to five decimal places; that is a property of the data, not of the design. The failure it sets up is the NAV sheet and the ledger/TD checks each holding a defensible but *different* "the price", with nothing to point at.

## What this does

Rather than teach the SQL the same list a second time, the list is **published**: `price_source_priority` is rewritten from `PRICE_FEEDS` on every application start. Java stays the source of truth, the table is derived, and the query reads it. **Change the order in `PRICE_FEEDS`, deploy, and the SQL follows** — which is the actual requirement; a second hardcoded list would just drift again.

- `V1_248__price_source_priority.sql` — table + seed of the current order, so it is usable before the app has started once.
- `PriceSourcePriorityPublisher` — `@EventListener(ApplicationReadyEvent)`, replaces the contents **wholesale in one transaction** (the `UNIQUE` constraint on `rank` would otherwise trip partway through a reordering).

**Price resolution behaviour is unchanged** — the publisher only writes the table. No existing code path reads it.

## Deliberately not the D5 pattern

`benchmark_category_proxy` was created, seeded, and then never read, which is why D5 exists in the centralization plan. The direction here is the opposite: the **code** stays authoritative and the table is a projection of it, so the two cannot fall out of step even if the table is ignored.

## Tests

`PriceSourcePriorityPublisherTest` — 3 tests, passing: every feed is written in `PRICE_FEEDS` order with ranks `1..n`; the table is cleared first so a removed source cannot linger; and the published order covers every `PriceSource` value with BlackRock/Morningstar/EODHD at the top, so adding an enum constant without placing it in `PRICE_FEEDS` fails the build rather than silently ranking 999 in SQL.

## Companion PR — merge/deploy order matters

[TulevaEE/tuleva#323](https://github.com/TulevaEE/tuleva/pull/323) switches the NAV price query to read this table. **This PR must deploy first** — until the table exists, that query errors.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/PriceSourcePriorityPublisher.java
+++ b/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/PriceSourcePriorityPublisher.java
@@ -1,0 +1,55 @@
+package ee.tuleva.onboarding.comparisons.fundvalue;
+
+import java.util.List;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+/**
+ * Publishes the {@link PriorityPriceProvider#priceFeeds()} ordering into {@code
+ * price_source_priority}, so SQL consumers rank price sources exactly the way this service does.
+ *
+ * <p>The Java list stays the single source of truth; the table is a derived copy, rewritten on
+ * every application start. The consumer it exists for is the NAV price query in the tuleva repo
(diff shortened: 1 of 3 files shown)


taavipertman-glitch:
Not wanted — the price-source ordering stays in the SQL query itself, matching PriorityPriceProvider.PRICE_FEEDS by hand. No table, no publisher, no cross-repo deploy dependency. Closing unmerged; nothing here was deployed.

PR #1873, opened 2026-08-19
Stop asking the register about a period the pillar cannot measure

stenehrlich-tuleva:
## What broke

`/portfolio` answered **400** for any start date after the beginning of April, with:

```json
{"errors":[{"code":"epis.message.exception"}]}
```

Captured in production on 2026-08-18. All-time returned 200 in 1.46 s; `?from=2026-05-01&to=2026-08-18` returned 400 in 438–932 ms. Not a timeout.

## Why

`ReturnsService.getRevisedFromTime` moves a second pillar start forward to the next transfer mandate fulfilment date. For `from=2026-05-01` the period ends 31.07.2026, so the fulfilment date is `nextWorkingDay(31.08)` = **01.09.2026** — after the requested `to` of 18.08 and in the future.

`AccountOverviewProvider` then clamps `endTime` up to match, and the register is asked for a cash flow statement dated 01.09.2026 → 01.09.2026. It refuses. `RestResponseErrorHandler` parses that 4xx body and rethrows it, so the code reaches the browser verbatim.

The all-time request makes the same lifetime cash flow call and succeeds, so the future-dated window is the only difference between the 200 and the 400.

## The change

`ReturnsService.hasMeasurablePeriod` answers whether the revised start still falls inside the period. `PortfolioService.annualReturnRate` checks it before asking, and reports no rate when it does not — the same thing it already does for a group with no rate. No provider runs, no account overview is built, no register call is made.

## Scope

Deliberately narrow. `/v1/returns` is untouched, so the fund comparison page keeps exactly the behaviour it has today. It reaches the same revision and has the same underlying defect; fixing it needs an answer to "what should a second pillar return mean for a period shorter than the mandate fulfilment cycle", which is a separate decision.

## Tests

- `ReturnsServiceSpec` — `hasMeasurablePeriod` across a February start (measurable, revised start 01.05.2026) and May/June/July/August starts (not measurable). The February row is the negative control: January to March produce fulfilment dates in April/May and do not exercise this at all.
- `PortfolioServiceTest` — no rate is requested and none is reported when the period cannot be measured.
- 269 tests green across the Portfolio, Returns and ReturnCalculator surface locally.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioController.java
+++ b/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioController.java
@@ -31,10 +31,18 @@ public class PortfolioController {
       @AuthenticationPrincipal AuthenticatedPerson person,
       @RequestParam(required = false) @DateTimeFormat(iso = DATE) @Nullable LocalDate from,
       @RequestParam(required = false) @DateTimeFormat(iso = DATE) @Nullable LocalDate to) {
-    var endDate = to == null ? LocalDate.now(clock) : to;
-    if (from != null && from.isAfter(endDate)) {
-      throw new ResponseStatusException(
-          BAD_REQUEST, "Invalid portfolio period: from=" + from + ", to=" + endDate);
+    var today = LocalDate.now(clock);
+    var endDate = to == null ? today : to;
+    if (from != null) {
+      if (from.isAfter(endDate)) {
+        throw new ResponseStatusException(
+            BAD_REQUEST, "Invalid portfolio period: from=" + from + ", to=" + endDate);
+      }
+      if (from.isAfter(today)) {
+        throw new ResponseStatusException(
+            BAD_REQUEST,
+            "Portfolio period starts in the future: from=" + from + ", today=" + today);
+      }
     }
     return portfolioService.getPortfolio(person, from, endDate);
   }
--- a/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioService.java
+++ b/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioService.java
@@ -155,6 +155,9 @@ public class PortfolioService {
 
   private Optional<BigDecimal> annualReturnRate(
       AuthenticatedPerson person, String key, LocalDate from, LocalDate to) {
+    if (!returnsService.hasMeasurablePeriod(from, to, List.of(key))) {
+      return Optional.empty();
+    }
     return returnsService.get(person, from, to, List.of(key)).getReturns().stream()
         .map(Returns.Return::getRate)
         .filter(Objects::nonNull)
--- a/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioService.java
+++ b/src/main/java/ee/tuleva/onboarding/account/portfolio/PortfolioService.java
@@ -17,6 +17,7 @@ import ee.tuleva.onboarding.fund.FundRepository;
 import ee.tuleva.onboarding.savings.fund.SavingsFundConfiguration;
 import ee.tuleva.onboarding.savings.fund.nav.FundNavProvider;
 import java.math.BigDecimal;
+import java.time.Clock;
 import java.time.LocalDate;
 import java.util.HashMap;
 import java.util.List;
@@ -45,6 +46,7 @@ public class PortfolioService {
   private final SavingsFundConfiguration savingsFundConfiguration;
   private final ReturnsService returnsService;
   private final FundNavProvider fundNavProvider;
+  private final Clock clock;
 
   public Portfolio getPortfolio(
       AuthenticatedPerson person, @Nullable LocalDate from, LocalDate to) {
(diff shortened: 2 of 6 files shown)


vootelerotov:
I pushed a third commit that reverts the first two and fixes the cause instead. Squash on merge (or rebase the middle two out) and the PR becomes +188/−8. Reasoning below, since reverting someone else's commits deserves an explanation rather than a diff.

### The revision is wrong on every row, not just the ones that 400

`getRevisedFromTime` for a personal-only request, run against the real code with a fixed clock at 2026-08-22:

| requested `from` | start actually used (P2) | `hasMeasurablePeriod`? |
|---|---|---|
| 2026-01-01 | **2026-05-01** | true |
| 2026-02-15 | **2026-05-01** | true |
| 2026-03-31 | **2026-05-01** | true |
| 2026-04-01 | **2026-09-01** | false |
| 2026-05-01 | **2026-09-01** | false |
| 2026-06-01 | **2026-09-01** | false |
| 2026-08-01 | **2027-01-01** | false |

The second-pillar start is never the date asked for. The April cutover is just where the wrongness crosses today and the register starts refusing.

`hasMeasurablePeriod` fires only on the bottom four rows. The top three are equally wrong and unchanged by it: "this year" returns an XIRR of May→August labelled January→August. If the market fell 10% in Feb–April and recovered 8% since May, the page shows roughly +8% to someone who is down on the year. `annualReturnRate` keeps only `Return::getRate` and discards `Return.from`, while `Portfolio.from` still reports the requested date, so nothing surfaces the discrepancy.

### Root cause

`SECOND_PILLAR` and `THIRD_PILLAR` are `Return` keys. No retriever ever writes them to `fund_value`, so `findEarliestDateForKey` is empty for them, `latestCommonStartDate` falls back to `.orElse(fromDate)`, and the guard `!fromDate.isAfter(latestCommonStartDate)` compares `fromDate` against itself — unconditionally true. The revision then fires on every personal request.

That `.orElse(fromDate)` was not designed for this. `b47d42b45` made a missing key **throw**; `86ca8cced` ("Hotfix") replaced the throw with the fallback 19 minutes later. Production immediately proved requests arrive with keys that have no NAV rows — the personal ones — and we have been building on that patch since. `c1112f896` patched the empty-keys edge, `0008f0463` fixed third-pillar-wins-when-present, and this PR is the third patch around the same mismatch.

### The fix

Five lines, purely additive. Nothing removed or reordered:

```java
private static final Set<String> PERSONAL_KEYS = Set.of(SECOND_PILLAR, THIRD_PILLAR);
...
if (hasKeys && PERSONAL_KEYS.containsAll(keys)) {
  return toInstant(fromDate);
}
```

The mandate revision is a fund-comparison concept — it aligns a hypothetical switch with the mandate cycle so a personal return is comparable to a simulated index or fund return. A personal-only request has nothing to align to. The person's units were in the fund from `from`, `ReturnCalculator` is pure XIRR over EPIS cash flows with no NAV dependency, and the register will answer the requested window.

Every other key set keeps today's behaviour, including a comparison fund whose NAV rows have not been backfilled yet.

### What this PR got right and I kept

Both ideas from the original commits survive, in places that do not depend on the revision:

- **A period that has not begun is refused**, in `PortfolioController` where the other date validation already lives.
- **A period with no length withholds its rate.** This concept was correct; it was measuring the revised start, which was itself the defect. Applied to the *requested* window it is `PortfolioService.hasMeasurableLength`, and it withholds only the rate — the valuation still renders, so `from=today&to=today` keeps working. That matters: `from == today` on a 100→101 balance otherwise hands XIRR a one-day span and it annualises to roughly 3678%.

### Review

The approach changed twice under adversarial review before landing, so it is not the first thing that occurred to me:

- The first regression test only exercised the package-private `getRevisedFromTime`. Now there is a service-boundary test through `get()` whose red output is `startTime=2026-09-01Z` against `endTime=2026-08-22Z` — the inverted window that produces `epis.message.exception`, pinned in a test.
- The rule was originally "no key has NAV history", which also caught an active fund awaiting backfill; a race there could turn a real rate into `ReturnCalculator`'s 0% sentinel. Narrowed to personal keys, and there is a two-row table pinning that a mixed key set still revises whether or not the comparison key has NAV history.
- The same-day annualisation above was found by review, not by me — I had wrongly assumed the solver would throw and yield null.

Full suite green: 6792 tests, 0 failures.

### Before merging

- **A published number will move.** The first-vs-second-pillar page requests `keys[]=SECOND_PILLAR` for 2018-01-01→2025-12-31, so its rate is now measured from January 2018 instead of ~May 2018. That is a correction — the page grows a 2018-01-01 snapshot by that rate, so the windows finally match — but the figure on screen changes.
- **The register is mocked in these tests.** They prove we stopped constructing the invalid window, not that EPIS answers the corrected one. Worth one manual `/v1/portfolio?from=<after April>` against staging.
- **One-day annualisation stays reachable** for a genuine 1-day window. Third pillar already behaves this way, so the pillars now agree rather than one being broken — but whether the UI should offer such short periods is a product question, left alone here.

Separately: `ReturnCalculator.getSimulatedRateOfReturn` returns `ZERO` when a comparison key has no NAV data, so "no data" renders as a real 0%. `Return.rate` is already `@Nullable` and the frontend filters `rate === null`. Out of scope here, worth its own ticket.

vootelerotov:
@stenehrlich-tuleva, let me know if you agree with the latest commit.

vootelerotov (commented):
LGTM!

vootelerotov (approved):

PR #1875, opened 2026-08-21
Drop carried-forward EODHD closes for Euronext Paris tickers

ErkoRisthein:
## Problem

On 2026-08-19 EODHD served **carried-forward (stale) closes** for all three Euronext Paris ETF listings (USAS.PA, WLSC.PA, GAGH.PA) — repeating the previous day's close instead of the official exchange close. The rows carried plausible non-zero volumes (USAS showed the real Euronext trade count), so the staleness is undetectable from the EODHD payload alone. This has happened repeatedly on low-activity Paris trading days.

Because `PriorityPriceProvider` ranks EODHD above Euronext at the same date, the stale row shadowed the correct exchange close and would have overstated the TKF100 NAV by ~0.05% (TKF100 holds 528,951 units of USAS; 5.007 stale vs 4.993 official). Manual intervention (deleting the rows in prod) was needed — see the Fund Value Integrity Check alerts from 2026-08-21.

## Fix

In `EODHDValueRetriever`, for `.PA.EODHD` tickers only:

1. Extend each fetch 7 days back to establish a baseline close before the requested range.
2. Treat a row whose adjusted close **exactly repeats the previous day's** as suspect.
3. Cross-check the suspect row against the **same-date Euronext official close already stored in `index_values`** (`ISIN.XPAR`):
   - Euronext agrees → genuine repeat → row kept
   - Euronext disagrees → confirmed stale → row skipped (warn log)
   - No same-date Euronext row yet (it is fetched after EODHD in the same indexing cycle) → skipped conservatively; the next hourly cycle refetches the date (the indexing job resumes from `min(latest across EODHD keys)+1`) and inserts it once confirmed — self-healing, at most one cycle of delay

While an EODHD row is skipped, price resolution falls through to the same-date Euronext row by date priority, so NAV correctness is never affected. XETRA, EUFUND, and forex series are untouched.

This would have caught all three stale rows from 2026-08-19 (a volume-based check would have caught none — EODHD reported fake non-zero volumes on the stale rows).

## Notes

- EODHD support confirmed and fixed the 2026-08-19 data after we reported it, but the fix only came after a support ticket — hence guarding at ingest.
- `index_values` is append-only, so a stale insert can never be corrected later; per the module's CLAUDE.md the retriever must not insert bad data in the first place.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/EODHDValueRetriever.java
+++ b/src/main/java/ee/tuleva/onboarding/comparisons/fundvalue/retrieval/EODHDValueRetriever.java
@@ -17,6 +17,7 @@ import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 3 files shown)


ErkoRisthein:
@codex review

PR #1876, opened 2026-08-21
Only check carried-forward closes for rows inside the requested range

ErkoRisthein:
Follow-up to #1875. The 7-day baseline rows fetched for Paris tickers exist only to give the repeat-detection a predecessor to compare against — they are filtered out before insert. But the carried-forward check still ran on them, so a historical divergence (WLSC.PA 2026-08-14: EODHD 5.085 vs Euronext 5.086) paged Sentry at error level on every hourly indexing run right after the deploy, even though nothing could or should be done about a week-old date.

Now rows before `startDate` pass through unchecked (they are dropped by the existing range filter anyway) and still serve as comparison baselines; the repeat check and its alerts fire only for dates that are actually candidates for insertion.


ErkoRisthein:
@codex review

PR #1890, opened 2026-08-24
Anchor the fee policy back-extension on the fund, not on its first row

### LINKED ISSUE #1887: Fee policy back-extension is anchored on the earliest row, not the fund's inception
## What

`FeeChargedToFundPolicy.Resolver.chargedOn` back-extends the earliest policy row to any date before it:

```java
if (predatesTheFoundingPolicy(date)) {
  return foundingPolicy().chargedToFund();
}
throw new IllegalStateException("Gap in the fee policy, no row covers this date: ...");
```

```java
private boolean predatesTheFoundingPolicy(LocalDate date) {
  return date.isBefore(foundingPolicy().validFrom());
}
```

The back-extension itself is deliberate and I am not proposing to remove it — `chargedToFund_readsDatesBeforeTheFundExistedAsTheFoundingPolicy` pins it, and a NAV recomputed past the inception date should not fail on the fee policy.

The problem is the anchor. It compares against the earliest **row**, not against the fund's inception, even though `TulevaFund` carries `inceptionDate`. So these two states are indistinguishable to the resolver:

1. a date genuinely before the fund existed — the case the back-extension is for;
2. a policy row whose `valid_from` is **later** than the fund's inception — a misconfiguration, where every day between inception and `valid_from` silently takes that row's answer instead of failing.

## Why it matters more than it looks

`FeeRepositoriesIntegrationTest.EveryFundAndFeeTypeHasAPolicy` reads as the guard against exactly case 2 — `datesToProbe` starts with `fund.getInceptionDate()` and the test asserts `chargedOn` does not throw:

```java
private List<LocalDate> datesToProbe(TulevaFund fund, FeeType feeType) {
  var dates = new ArrayList<LocalDate>();
  dates.add(fund.getInceptionDate());
  ...
```

But with the back-extension in place that probe **cannot** throw for a late-starting row. The test would stay green on the misconfiguration it appears to be written against, so it is asserting less than it looks like it asserts.

Nothing is wrong in production: `V1_243` seeds all eight `fund × feeType` rows at each fund's inception (`TUK75`/`TUK00` 2017-03-28, `TUV100` 2019-10-15, `TKF100` 2026-02-02), and every row is open-ended. This is about the guard, not a live defect.

## Options

Either would close it; the second is the smaller change.

1. **Anchor the back-extension on inception** — `date.isBefore(fund.getInceptionDate())` instead of `foundingPolicy().validFrom()`, so a row that starts after inception leaves a real gap and throws like any other.
2. **Assert the invariant directly in the test** — `min(valid_from) <= fund.getInceptionDate()` per fund and fee type, rather than inferring it from a non-throw.

Found while reconciling `work/investeerimistegevus/docs/Investment Data Sources Guide.md` against the deployed code (TulevaEE/tuleva#291), which claimed the policy has no fallback at all. That doc sentence is being corrected separately to describe the back-extension.

taavipertman-glitch:
Closes #1887.

## What

`FeeChargedToFundPolicy.Resolver.chargedOn` answers a date before the fund existed with the founding row rather than throwing, so a NAV recomputed past the inception date does not fail on the fee policy. **That is deliberate and stays** — `chargedToFund_readsDatesBeforeTheFundExistedAsTheFoundingPolicy` pins it.

The problem was the anchor:

```java
private boolean predatesTheFoundingPolicy(LocalDate date) {
  return date.isBefore(foundingPolicy().validFrom());   // the row, not the fund
}
```

Anchored on the earliest row, the row back-extends over **its own late start**. So two different states were indistinguishable: a date genuinely before the fund existed, and a policy row that begins *after* the fund did — where every day in between was silently answered instead of reported as the gap it is.

The `Resolver` already carries `fund`, so the fix is to ask the fund.

## It also gives the guard test something to catch

`FeeRepositoriesIntegrationTest.EveryFundAndFeeTypeHasAPolicy` reads as the guard against exactly that misconfiguration — `datesToProbe` starts with `fund.getInceptionDate()` and asserts `chargedOn` does not throw. Under the old anchor that probe **could not** throw for a late-starting row, so the test asserted less than it appeared to. Now it holds the line it reads as holding.

## Tests

New `chargedToFund_throwsWhenTheFirstRowStartsAfterTheFundDid`, verified red before the fix:

```
FeeChargedToFundPolicyTest > chargedToFund_throwsWhenTheFirstRowStartsAfterTheFundDid() FAILED
    Expecting code to raise a throwable.
```

`FeeChargedToFundPolicyTest` and `FeeRepositoriesIntegrationTest` green.

## No behaviour change in production

`V1_243` seeds all eight `fund × feeType` rows at each fund's inception (`TUK75`/`TUK00` 2017-03-28, `TUV100` 2019-10-15, `TKF100` 2026-02-02), every one open-ended. Nothing resolves through the changed branch today.

Found while reconciling `work/investeerimistegevus/docs/Investment Data Sources Guide.md` against deployed code — TulevaEE/tuleva#291 claimed the policy has no fallback at all, and that sentence is corrected separately.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/fees/FeeChargedToFundPolicy.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/fees/FeeChargedToFundPolicy.java
@@ -63,7 +63,7 @@ public class FeeChargedToFundPolicy {
       if (applicable.size() == 1) {
         return applicable.getFirst().chargedToFund();
       }
-      if (predatesTheFoundingPolicy(date)) {
+      if (predatesTheFund(date)) {
         return foundingPolicy().chargedToFund();
       }
       throw new IllegalStateException(
@@ -75,8 +75,15 @@ public class FeeChargedToFundPolicy {
               + date);
     }
 
-    private boolean predatesTheFoundingPolicy(LocalDate date) {
-      return date.isBefore(foundingPolicy().validFrom());
+    /**
+     * A date before the fund existed is answered with the founding row rather than thrown on, so
+     * that a NAV recomputed past the inception date does not fail on the fee policy. The anchor is
+     * the fund's inception and not the earliest row's {@code valid_from}: anchoring on the row
+     * would make it back-extend over its own late start, so a policy that begins after the fund did
+     * would be silently answered for the days in between instead of reported as the gap it is.
+     */
+    private boolean predatesTheFund(LocalDate date) {
+      return date.isBefore(fund.getInceptionDate());
     }
 
     private Policy foundingPolicy() {
--- a/src/main/java/ee/tuleva/onboarding/investment/fees/FeeChargedToFundPolicy.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/fees/FeeChargedToFundPolicy.java
@@ -75,13 +75,6 @@ public class FeeChargedToFundPolicy {
               + date);
     }
 
-    /**
-     * A date before the fund existed is answered with the founding row rather than thrown on, so
-     * that a NAV recomputed past the inception date does not fail on the fee policy. The anchor is
-     * the fund's inception and not the earliest row's {@code valid_from}: anchoring on the row
-     * would make it back-extend over its own late start, so a policy that begins after the fund did
-     * would be silently answered for the days in between instead of reported as the gap it is.
-     */
     private boolean predatesTheFund(LocalDate date) {
       return date.isBefore(fund.getInceptionDate());
     }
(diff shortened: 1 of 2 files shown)


taavipertman-glitch:
Superseded by #1936, which includes both commits from this PR unchanged.

Folded in rather than merged separately because #1897 extends the same `FeeChargedToFundPolicy.Resolver` and the two conflict on adjacent lines. Together they also interact: once `chargedOn` is asked about every accrual date, anchoring the back-extension on the fund's inception rather than the first policy row is what stops a month reaching back past that row from throwing.

Branch `fix/fee-policy-inception-anchor` is left in place.

PR #1894, opened 2026-08-25
Split the savings fund tax report by investment account

stenehrlich-tuleva:
## What

Under TuMS § 17², units bought with money from a declared investment account are taxed under that regime: the gain on them is not declared as a gain, and tax falls due only once more has left the account than went into it. Units bought with ordinary money are declared the usual way. Someone holding both holds two pools with separate acquisition costs, and a single FIFO or weighted average over all of it answers neither.

Three steps: a person can tell us which account is theirs; every savings fund transaction now carries the account on the other side of it; the report splits by that and runs the existing calculator once per pool.

## Why the calculator is untouched

`SavingsFundCostBasisCalculator` was already a pure function of a transaction list, so the pools are two calls to it rather than a second algorithm.

## Where the account on the other side comes from

No new bank data and no ledger change. `IssuerService` already records the `FUND_SUBSCRIPTION` ledger transaction with the payment's id as its external reference, so a subscription reaches `SavingFundPayment.remitterIban`; redemption entries already carry `REDEMPTION_REQUEST_ID`, so a redemption reaches `RedemptionRequest.customerIban`. One query per direction, both indexed — subscriptions via `findPayments(partyId)` on `idx_saving_fund_payment_party`, the declaration by primary key.

Note `SavingsFundLedger.MetadataKey.PAYER_IBAN` is declared and never written anywhere. It looks like the natural place for this and is not.

## What it refuses to answer

A redemption paid to an account that did not buy the units cannot be assigned to a pool without deciding whether the deferral under § 17² survived — and that decision changes someone's tax return. Same for a transaction whose counterparty we never recorded: that is missing evidence, not evidence of ordinary treatment. In both cases the report returns the unsplit numbers with a null investment-account total, and the page says it could not split them. **The wording of that case needs compliance sign-off before this is announced to anyone.**

## Known and deliberate

- Declaring an account reclassifies all history from it, including payments made before it was an investment account. Simplest thing that works; the copy does not yet say so.
- Nothing checks that the declared account belongs to the person. A parent acting for a child can name their own. `IbanWhitelistService` already knows which accounts have paid in, if we want to close that.
- Taking a declaration back is not exposed. Declaring another account replaces the previous one.

## Merge order

Merge before onboarding-client#1626, which reads `investmentAccount` off the report.

### DIFF
--- a/.githooks/pii-iban-allow
+++ b/.githooks/pii-iban-allow
@@ -6,6 +6,11 @@
 # Never add a person's account, and never add a live pension fund account -
 # those belong in the gitignored terraform tfvars, not in code.
 # One IBAN per line; anything else is ignored.
+#
+# Repo-invented savings-fund test fixtures, long public in this repository's
+# test files (Juri Tamm demo persona's account and the deposit-account fixture):
+EE157700771001802057
+EE442200221092874625
 EE861010220306591229
 EE651010220306497226
 EE241010220306719221
--- a/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
+++ b/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
@@ -25,7 +25,8 @@ public record Transaction(
     String isin,
     CashFlow.Type type,
     BigDecimal units,
-    BigDecimal nav)
+    BigDecimal nav,
+    String counterpartyIban)
     implements Comparable<Transaction> {
 
   public Transaction {
--- a/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
+++ b/src/main/java/ee/tuleva/onboarding/account/transaction/Transaction.java
@@ -25,8 +25,7 @@ public record Transaction(
     String isin,
     CashFlow.Type type,
     BigDecimal units,
-    BigDecimal nav,
-    String counterpartyIban)
+    BigDecimal nav)
     implements Comparable<Transaction> {
 
   public Transaction {
--- a/src/main/java/ee/tuleva/onboarding/capital/transfer/iban/IbanValidator.java
+++ b/src/main/java/ee/tuleva/onboarding/capital/transfer/iban/IbanValidator.java
@@ -143,7 +143,8 @@ public class IbanValidator implements ConstraintValidator<ValidIban, String> {
       }
     }
     boolean ok = mod == 1;
-    if (!ok && log.isDebugEnabled()) log.debug("IBAN failed checksum: {}", iban);
+    if (!ok && log.isDebugEnabled())
+      log.debug("IBAN failed checksum: country={}, length={}", cc, len);
     return ok;
   }
 }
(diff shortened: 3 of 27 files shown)


vootelerotov:
@stenehrlich-tuleva, did some changes here. Let me know if you are happy with them.

stenehrlich-tuleva:
Read all eight commits — thank you, several of these are mine to own.

The counterparty account on the shared `Transaction` record is the one I should have caught: that record is what `GET /v1/transactions` serialises, so I had put bank account numbers onto an API. `TransactionsWithCounterparties` keeping it inside the savings fund module is right, and `keepsTheCounterpartyAccountOutOfTheTransactionList` is the test that would have stopped me. Worth saying out loud that `pii-scan` cannot catch this class of thing — it reads literals in the diff, not what the code puts on the wire.

The closed-year one stings: I bounded the balance check to the report's end date and left the unknown-counterparty check running over all history, one line away. Same defect, didn't look at its neighbour.

Same-instant ordering and the concurrent declaration I had both seen and judged not worth the code. You were right on both counts, and `ACQUISITIONS_FIRST_WITHIN_AN_INSTANT` reads better than the comparator I would have written. Keeping the account number out of the checksum log is a good catch on pre-existing code that my feature made reachable with a person's own account.

On validation living in two places — agreed, and I should explain why it ended up in one. CI's `pii-scan` refuses a checksum-valid IBAN on an added line, so validating in the service meant every test had to hand it a real-looking account number. Moving it to `@ValidIban` on the command was how the fixtures became numbers that are obviously nobody's. Your `IllegalArgumentException` at the storage boundary restores the half I dropped; the one test that still needs a real IBAN uses an account already on `.githooks/pii-iban-allow`.

**On the red build.** `ci/circleci: build` failed on `da9231ee` with two failures, both in `SavingFundPaymentUpsertionServiceIntegrationTest`:

- `endToEnd_processesXmlAndStoresPaymentsInDatabase()`
- `outgoingPaymentToInvestmentAccountCreatesLedgerEntry()`

Both `Expected size: 1 but was: 2`, a leftover `saving_fund_payment` row (`partyId=null`, remitter `Mari Maksja`). I do not think it is either of us:

- the class is bank-XML payment upsertion, nothing on the split's path
- it passes alone, and passes for the whole `savings.fund` package
- it only fails in the full 6949-test run, which is the shared `mem:testdb` shape
- "expected 1, got 2" is a leftover row rather than a logic error

The thing that argues the other way is that my head was green and yours is not, and the new `@DataJpaTest` classes could have shifted suite ordering enough to expose pollution that was already there. A re-run would settle it — could you kick one off? I do not have a CircleCI token.

**And the question I actually need answered:** do these eight commits count as your review, or are you still going through it? Asking because you tend to push rather than click, and I do not want to sit waiting on an approval that already happened.

Merge order stays: this one first, then onboarding-client#1626, which reads `investmentAccount` off the report. That one is green and I have not touched it since you started.

stenehrlich-tuleva:
Follow-up on the red build: the full suite passes on `da9231ee` locally. All 6949, no failures, 25m30s.

So the ordering-shift theory I raised against myself does not hold either — nothing in the eight commits reproduces it off CI. That leaves the CI run itself, and a re-run should be green.

vootelerotov (approved):
LGTM!

PR #1897, opened 2026-08-25
Apply the fee policy per accrual date, not once per month

### LINKED ISSUE #1856: Apply the fee policy per accrual date on the three NAV-facing readers
#1847 established that whether a fee reaches the fund is a **per-date** question, and applied it that
way in `PeriodicTdAttributionService.computeFeeDragPeriod` and `LedgerAccrualConsistencyChecker`.
Three NAV-facing readers still gate a **multi-day sum** on a **single day's** answer.

| Path | What it sums | Policy evaluated at |
|---|---|---|
| `NavCalculationService.navFacingAccrual` | `getUnsettledAccrual` — the whole unsettled fee month | `positionReportDate` |
| `FeeAccrualPositionSyncJob.chargedAccrual` | `getUnsettledAccrual` | `navDate` |
| `TransactionInputService.getAccruedFees` | `getAccruedFeesForMonth` | `asOfDate` |

`calculateFeesForNav` ends in `getUnsettledAccrual(fund, feeType, positionReportDate)`, which sums
`daily_amount_gross` over `fee_month = positionReportDate.withDayOfMonth(1)` and
`accrual_date <= positionReportDate`. `navFacingAccrual` then keeps or zeroes **all** of it on one
boolean.

### Why it matters

`recordDailyFees` posts to the ledger per accrual date. So on a fee month where the policy flips:

- the **ledger** holds entries only for the charged days,
- the **NAV liability**, the fee accrual **position** and the **investable cash** deduction are
  all-or-nothing for the whole month.

`LedgerAccrualConsistencyChecker` does resolve per date, so it would raise a FAIL for that fee on
every day of the month — a red daily check caused by the NAV and the ledger disagreeing, not by a
real divergence.

### Not reachable today

No fund has ever flipped, and flipping is a deliberate act (close a row, open a successor). The
seeded policy is `DEPOT = false` / `MANAGEMENT = true` for all four funds from inception. This is
the same shape of latent defect #1847 chose to fix on the TD side; it was missed on the NAV side.

### Why it was not fixed in #1847

Every clean fix is disproportionate to a latent defect in a PR that is green and carries staged
migrations:

- Changing the shape of `getAccruedFeesForMonth` churns ~40 mock call sites in
  `TransactionInputServiceTest` alone.
- Keeping a "constant over the window" fast path alongside a per-date slow path puts two code paths
  into NAV fee logic that can disagree — the bug class being fixed.
- Pushing the policy into SQL as a join on `investment_fee_policy` loses the gap/overlap/missing-row
  validation that `FeeChargedToFundPolicy` exists to enforce.

### What to do

- [ ] Give `FeeAccrualRepository` per-date variants of both sums returning raw `daily_amount_gross`
      (note `getUnsettledAccrual` rounds the **sum** to 2 dp, so the caller must round after summing,
      not per day, or the figure moves).
- [ ] Sum only the dates `FeeChargedToFundPolicy.Resolver.chargedOn` accepts, in one shared seam
      rather than three call sites.
- [ ] Delete `NavCalculationService.navFacingAccrual` — with `calculateFeesForNav` returning an
      already-filtered figure there is nothing left for it to gate, and one less place to apply the
      policy twice or not at all.
- [ ] Test: a fund whose DEPOT policy flips mid-month; assert the NAV fee liability equals the
      DEPOT ledger balance for that month, and that `LedgerAccrualConsistencyChecker` stays green.

Context: #1847, `FeeChargedToFundPolicy`, `V1_243__investment_fee_policy.sql`.

taavipertman-glitch:
Closes #1856. A month-to-date accrual is a sum over many days, but all three
NAV-facing readers asked "is this fee charged to the fund?" **once**, for a
single day, and applied that answer to the whole sum. A policy that flips
mid-month would then put NAV and the ledger out of step for that month and turn
`LedgerAccrualConsistencyChecker` red.

Concretely: Tuleva bears TUK75's depot fee, the board moves it to the fund on
15 September. Days 1–14 belong to Tuleva, 15–30 to the fund. Asking once for the
30th admits the **entire** month into NAV; asking once for the 1st admits none
of it. Neither is right.

**The fix is one seam, not three.** All three readers bottom out in the same two
queries, so the per-date split goes there:

- `FeeAccrualRepository.getAccruedFeesByDateForMonth` / `getUnsettledAccrualByDate`
  return the same rows grouped by `accrual_date` instead of pre-summed.
- `FeeChargedToFundPolicy.Resolver.sumChargedDays(...)` sums only the days the
  policy actually charges — reusing the resolver that already loads the policy
  rows once, so this costs no extra queries.

Then: `FeeCalculationService` (which feeds `NavCalculationService`),
`TransactionInputService.getAccruedFees` and
`FeeAccrualPositionSyncJob.chargedAccrual` all filter per day.

**`NavCalculationService.navFacingAccrual` is deleted, deliberately.** With the
policy applied upstream per date, re-gating the summed figure on one day's
answer would undo the fix — it would zero or admit a flipped month whole. That
also means `FeeResult` is now the NAV-facing accrual rather than the raw one;
the accrual rows themselves are unchanged and still recorded in full, which the
integration test now asserts explicitly.

**Gap and overlap validation is preserved and strengthened.** `chargedOn` still
throws on an uncovered or doubly-covered date — and now it is asked about every
accrual date, not just one, so a gap in the middle of a month is caught instead
of skipped. `sumChargedDays_stillRefusesAGapRatherThanSkippingTheDay` pins that
a day no row covers is an error, not a silent zero.

Scale is preserved exactly: the filtered sum rounds to 2 like the
`ROUND(SUM(...), 2)` it replaces, and returns a plain `0` for no rows, so no
stored figure changes shape.

**Tests.** `sumChargedDays_splitsAMonthThePolicyFlipsInsteadOfTakingOneDaysAnswer`
is the case this exists for: policy false to 14.09, true from 15.09, and only
the post-flip days are summed. The mock churn #1847 predicted was real — ~40
call sites across four suites — but is mechanical, and every behavioural
assertion is unchanged except the three that pinned the old whole-month gating.
`NavPipelineIntegrationTest`'s settlement test now seeds the policy it depends
on rather than relying on a default that made its depot figures a correct zero.

Full `./gradlew test` green.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Closes #1856.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/fees/FeeAccrualRepository.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/fees/FeeAccrualRepository.java
@@ -1,9 +1,12 @@
 package ee.tuleva.onboarding.investment.fees;
 
+import static java.util.stream.Collectors.toMap;
+
 import ee.tuleva.onboarding.fund.TulevaFund;
 import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import lombok.RequiredArgsConstructor;
 import org.springframework.jdbc.core.simple.JdbcClient;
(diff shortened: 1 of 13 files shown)


taavipertman-glitch:
Superseded by #1936, which carries this change re-applied onto master.

Master has since added the `NavFees` port/adapter, so this could not be merged as-is: the fix had to be re-expressed behind that ACL rather than importing `investment.fees` types into `NavCalculationService`. #1936 also folds in #1890 — the two edit the same `Resolver`, and making `chargedOn` run over every accrual date turns #1890's inception anchor from cosmetic into load-bearing.

The bug is unchanged and still live on master. Branch `fix/fee-policy-per-accrual-date` is left in place.

PR #1900, opened 2026-08-25
Send third pillar payment arrived emails daily

ErkoRisthein:
Flips the payment-arrived job to production mode: daily at 12:00 Tallinn, dry-run off, capped at 200 recipients per run (the guard refuses to run above the cap rather than truncating).

Today's dry run reported an audience of 142, so the first real run clears the backlog within the cap. The nightly 02:00 transaction sync feeds the source table, and the already-emailed guard makes daily runs idempotent. Staging and dev keep the weekly dry-run defaults.


ErkoRisthein:
Superseded by the consolidated rollout PR.

PR #1901, opened 2026-08-25
Deduplicate the cross-sell email nudges and clarify the hasTulevaUser flag

ErkoRisthein:
Extracts the second-pillar, payment-rate, and membership suggestion blocks (copied into 13 templates, drifted: 0,3% vs 0,5% fee threshold, differing emphasis, a stale campaign button, one broken double-scheme link) into one partial per family and language — 353 duplicated lines become 32 includes with canonical wording. Only the 13 affected templates change bytes; the other 39 stay identical and will not republish.

Also renames the ambiguous hasAccount merge variable to hasTulevaUser (it checks for a users row, i.e. a Tuleva web login).

Merge timing: the template publish is instant on merge while the ECS deploy lags ~20 minutes, and the payment-arrived job sends at 12:00 — merge outside the noon window so the renamed merge tag and claim switch together.

### DIFF
--- a/emails/fixtures/third_pillar_payment_arrived_en.json
+++ b/emails/fixtures/third_pillar_payment_arrived_en.json
@@ -4,34 +4,34 @@
       "FNAME": "Mari",
       "amount": "300.00",
       "paymentDate": "12.08.2026",
-      "hasAccount": true,
+      "hasTulevaUser": true,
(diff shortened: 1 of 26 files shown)


ErkoRisthein:
Superseded by the consolidated rollout PR.

PR #1904, opened 2026-08-26
Remind savers who opened a savings fund account but never paid

PirjeK:
Closes the third trigger in the order recommended on TulevaEE/tuleva#390: the account is open, no payment ever arrived, and nobody follows up. Content side of TulevaEE/tuleva#373 for the adult segment. Mapping: TulevaEE/tuleva#371.

## What it does

A daily job at 12:00 Tallinn time picks savers whose own savings fund account was completed between 30 and 7 days ago, who have never sent us a payment and who have not been reminded before, and sends `savings_fund_first_payment_reminder_person_{et,en}` through Mandrill.

Recipients are capped at 200 per run. Above the cap the job sends nothing and logs an error — the same safety valve `AutoEmailSender` uses. That means a large backlog on the first run blocks itself on purpose; the count should be checked before this goes live.

## Who is left out

- anyone with a payment row of any status, including cancelled and returned
- minors, by birth date from the personal code rather than by the existence of a parent link
- people whose own account exists only because a company onboarding needed their KYC, when that company has paid
- anyone already reminded, tracked through the `email` table on the new `EmailType`

The last three rules come from the manual campaign in `tuleva` `knowledge/campaigns/mailchimp/drafts/tkf-maksmata-2026-08/`, which solved the same segmentation by hand.

## Why the onboarding table changed

`savings_fund_onboarding` only knew when the application started, not when the account opened. `V1_251` adds `status_changed_at` and status changes now record it. Existing rows fall back to the row creation time, which under-sends slightly on the first runs rather than waking up an old tail.

## Language

The job has no request locale, so the language comes from the saver's own preference in the latest `unit_owner` snapshot, defaulting to Estonian. That is why the English template is in here too.

## Templates

The nine Estonian templates in the first commit have no code path yet — they cover triggers still to be built. Publishing them is safe on its own: a template only starts sending once `EmailType` references its name.

## Worth a careful look

- migration number `V1_251` may need renumbering if another PR lands first
- the module boundary: `savings.fund.reminder` reaches into `mandate.email.persistence` and `notification.email`, the same way `AutoEmailSender` does
- the recipient query is raw SQL across six tables


PirjeK:
**Do not merge before 10 September 2026.** Marking this a draft until then.

On 11 August a manual Mailchimp campaign went to 630 adults with an open savings fund account and no payment — the same message this job automates ([analysis](https://github.com/TulevaEE/tuleva/tree/main/knowledge/campaigns/mailchimp/analysis/tkf-maksmata-2026-08)).

Anyone in that campaign who opened their account shortly before 11 August and still has not paid is inside this job's 7–30 day window, and would receive essentially the same letter a second time. The email log cannot prevent it: a campaign sent from the Mailchimp UI leaves no row in the `email` table.

From 10 September every account opened before 11 August has aged out of the 30-day window, so the overlap disappears on its own and no exclusion logic is needed. Merging earlier would need one.

Nothing else is blocking: content is approved and the tests are green.

PirjeK:
**No longer date-gated — ready for review.**

The duplicate risk described above is now handled in code rather than by waiting. `FirstPaymentReminderJob` never looks back past **12 August 2026**, whatever its 30-day window says, so nobody who received the 11 August campaign can be reminded again.

A Metabase count on 27 August put the overlap at **12 of the ~24 people** the first run would reach — half the first send would have been a repeat.

Handling it in code fixes the other end too: waiting for the window to clear on its own would have silently dropped the mid-August accounts if review had slipped past mid-September, because they age out of the 30-day window.

The constant is deliberately temporary and can go once every account it covers has aged out.

PirjeK:
**Status: content is signed off, code is green, waiting on an approving review.**

### What is in here

A saver opens a savings fund account, never pays in, and today hears nothing. This adds the letter and the job that sends it: 20 commits, 32 files, the Estonian and English templates, `EmailType`, the query, the sender, the daily job, and migration `V1_252` adding `status_changed_at` — `created_at` records when the application started, not when the account opened.

Nine Estonian templates for triggers that still have no code path ride along. They send nothing until something references their names.

### What changed since this was opened

- Karoliina reviewed the copy: non-breaking space in the fee, the fund called by its registered English name, the English letter pointing at English articles rather than Estonian ones
- Pirje approved the wording and the subject line
- A cutoff so the job never looks back past 12 August 2026, because a manual campaign reached 630 savers with the same message on 11 August and a Metabase count put 12 of the ~24 people the first run would reach inside it
- Erko renumbered the migration past master's `V1_251`, made the fee a merge variable, and dropped the company completion template as superseded by the live onboarded email

### What a reviewer should look at

Four judgement calls, all deliberate, none obvious:

1. **The cutoff constant.** A hardcoded date in the job, temporary by design, removable once every account it covers has aged out. The alternative was waiting until 11 September to merge, which would have silently dropped the mid-August accounts if review slipped.
2. **Language comes from the `unit_owner` snapshot**, defaulting to Estonian. A scheduled job has no request locale, and the analytics repositories behind `AutoEmailSender` already segment on this table.
3. **Minors are excluded by personal code**, not by the parent-child link alone, so a child account with no link recorded cannot receive an adult letter.
4. **Board members of a company that has paid are excluded**, so someone whose company funded its account is not told they have never paid.

### After this

#1919 is stacked on this branch: the same letter for child accounts, the recipient cap moved per segment and tightened to 50, and a Slack alert when a cap trips. It is easier to read after this one.

Expected first run: around 12 letters. Steady state: one or two a day.

PR #1905, opened 2026-08-27
Give the investment account migration a version of its own

taavipertman-glitch:
## Why

`origin/master` carries **two migrations numbered V1_249**:

| File | Landed | PR |
|---|---|---|
| `V1_249__constrain_instrument_reference.sql` | 25.08 13:26 | instrument reference constraints |
| `V1_249__create_investment_account.sql` | 26.08 13:55 | #1894 |

Flyway resolves every migration on the classpath *before* applying any of them, so it fails at startup:

```
FlywayException: Found more than one migration with version 1.249
```

**No instance has been able to start since the second one landed** — every commit merged since is sitting undeployed while ECS keeps serving the previous task, and `/actuator/health` stays UP throughout. Every Spring-context test fails with it too; the GitHub "Code Quality" workflow does not catch it.

Each PR was numbered correctly against master when it was written. The second was not re-checked after the first merged — the race CLAUDE.md describes for V1_198–V1_202.

## What

`V1_249__create_investment_account.sql` → **`V1_251`**. V1_250 is taken by the Java migration `V1_250__reference_data_history_trigger`.

Renumbering this one is safe: nothing has started since it existed, so it cannot have been applied anywhere. `V1_249__constrain_instrument_reference.sql` keeps its version because it *has* been applied.

## Test

File rename only, no content change. Verified locally that the Spring context loads and Flyway runs clean with the rename in place; without it every context test dies at `flywayInitializer`.


ErkoRisthein:
Fixed in https://github.com/TulevaEE/onboarding-service/commit/81db5b823103c3089b4cc7b4e284020debcd59e1

PR #1906, opened 2026-08-27
Stop H2 choking on the instrument reference check constraints

taavipertman-glitch:
## What breaks

`V1_249` added three CHECK constraints with `ALTER TABLE … ADD CONSTRAINT`. H2 cannot evaluate that form. Once a **second** Spring context opens the shared in-memory database, every `INSERT` and `UPDATE` on the constrained table fails:

```
DataIntegrityViolationException: SQL [UPDATE instrument_reference SET seb_position_name = ? WHERE isin = ?];
Check constraint invalid: "instrument_reference_instrument_type_check: " [23514-240]
```

Note the details: error **23514** ("the expression threw"), not 23513 ("the row violated it"); the constraint expression after the colon renders **empty**; and the failing statement sets `seb_position_name`, a column that check does not mention. The values are irrelevant — an `INSERT … VALUES ('ETF', 'equity', …)` fails the same way.

It reproduces **only in the full suite**, never when a test class runs alone, and **PostgreSQL is unaffected** — so the `pg` and `ci` profiles stay green while local `./gradlew test` goes red on tests that have nothing to do with instrument data. This is a documented H2 defect in this repo; a `CHECK` written inline inside `CREATE TABLE` is fine, only the `ALTER`-added form breaks.

## What this does

The constraints are what V1_249 was **for** — after the enum cutover an instrument change is an `UPDATE` typed into a console, and a mistyped instrument type has to fail at the moment of the change. So they stay in every environment the application runs in, and are dropped for **H2 only**, in `db/h2/V1_249_1`.

`InstrumentReferenceRulesTest` then asserts the same four rules over the rows, so nothing is left unchecked while the suite runs on H2:

- every `instrument_type` is `ETF` or `FUND`
- every `asset_class` is `equity` or `bond`
- every `benchmark_category_proxy` row names exactly one index target
- every `benchmark_category` on an instrument has a proxy

plus a regression test that an existing instrument can simply be updated.

`benchmark_category_proxy_index_target_check` is dropped alongside the two on `instrument_reference`: it is the same form and has escaped so far only because no test in the suite writes to that table.

## Test

- Full suite **without** the shim: exactly one failure, `anExistingInstrumentCanBeUpdated`, with the signature above. The four rule assertions pass — the data is fine, the evaluation is not.
- Full suite **with** the shim: green.

> Depends on #1905 — master currently has two migrations numbered V1_249, so no Spring context starts until that lands. This branch is cut from master and carries only the two new files.

### DIFF
--- a/src/test/groovy/ee/tuleva/onboarding/instrument/InstrumentReferenceRulesTest.java
+++ b/src/test/groovy/ee/tuleva/onboarding/instrument/InstrumentReferenceRulesTest.java
@@ -1,0 +1,89 @@
+package ee.tuleva.onboarding.instrument;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.jdbc.core.simple.JdbcClient;
+import org.springframework.transaction.annotation.Transactional;
+
+@SpringBootTest
+@Transactional
+class InstrumentReferenceRulesTest {
+
+  private static final List<String> INSTRUMENT_TYPES = List.of("ETF", "FUND");
+  private static final List<String> ASSET_CLASSES = List.of("equity", "bond");
+
+  @Autowired private JdbcClient jdbcClient;
+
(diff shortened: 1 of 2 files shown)


taavipertman-glitch:
Folded into #1907 — the H2 shim is what makes that PR's own tests runnable, so they review and merge as one change. The commit is carried over unchanged.

PR #1907, opened 2026-08-27
Split Xtrackers into its two issuers

taavipertman-glitch:
> **Depends on #1905** — master has two migrations numbered V1_249, so no instance starts at all until that lands.
>
> Note for the reviewer: this changes what the limit check reports. TKF100's Xtrackers pair stops being one 21% group and becomes 19% + 2%, so the alert that has fired every day since 25.08 stops firing on merge.

## Why

The provider limit caps what one **issuer of units** may hold, not what one management company may run. TKF100 holds `IE00BJZ2DC62`, issued by Xtrackers (IE) plc, and `LU0476289540`, issued by the Luxembourg SICAV — two legal issuers, one manager (DWS Investment S.A.).

Checked as one group they are 19,00% + 2,00% = **21,00%** against a 20,00% hard limit, which is what has been alerting daily since 25.08. As two they are 19,00% and 2,00%.

## What

- `Provider` gains `XTRACKERS_IE(IRELAND)` and `XTRACKERS_LU(LUXEMBOURG)`, and **keeps `XTRACKERS`** so the rows of earlier model versions still read.
- `V1_252` relabels **the version in force only**: allocations and position limits go to `XTRACKERS_IE`, except `LU0476289540` → `XTRACKERS_LU`; the provider limit row is renamed to `XTRACKERS_IE` and a `XTRACKERS_LU` row is added with the same percentages. TKF100 goes **6 → 7 rows**.

Earlier versions keep `XTRACKERS`: they record what was approved and what the check said on those days.

**The allocation rows are the load-bearing half** — `LimitCheckService` builds the ISIN → provider map from them, so the limit row alone would change nothing. And an issuer with no limit row is dropped from the check *without a word*, so the two halves must not exist apart: that is why this is one migration in one release rather than two hand-run statements.

**Settlement gets fixed on the way past.** `Provider` carries a `Domicile` that feeds `SettlementDateCalculator`'s holiday calendar, and `XTRACKERS(IRELAND)` is wrong for a Luxembourg fund.

## Second commit: H2 and the V1_249 check constraints

Unrelated to the split, included because it is a landmine for anyone writing a test against instrument data.

`V1_249` added three CHECK constraints with `ALTER TABLE … ADD CONSTRAINT`. H2 cannot evaluate that form: once a second Spring context opens the shared in-memory database, every `INSERT`/`UPDATE` on the constrained table fails with `Check constraint invalid: "…: " [23514-240]` — the expression threw, the constraint renders empty, and it happens on statements setting unrelated columns whatever values they carry. Full suite only, never a single class, and PostgreSQL is unaffected, so `pg`/`ci` stay green while local runs go red.

The constraints stay everywhere the application runs and are dropped for **H2 only** (`db/h2/V1_249_1`); `InstrumentReferenceRulesTest` asserts the same rules over the rows instead.

## Not changed

Nothing about a provider limit blocks anything, and this keeps it that way: a breach is a Slack line, `LimitCheckJob` only notifies. `LimitCheckService`, `ProviderLimitChecker`, `ProviderLimit` and `InstrumentDataValidator` are untouched.

When the prospectus clause goes (the register marks A2 `kaotamisel`, next prospectus version effective 18.09.2026), the check is deleted with it — `ProviderLimitChecker`, `ProviderBreach`, the `CheckType.PROVIDER` wiring and the rows. The two enum values stay, because the domicile they carry is a real fact the settlement calendar reads.

## Test

- Unit: two ISINs under two issuers are checked separately; the same fixture under one provider is 21% and HARD.
- Integration: the whole check run for TKF100 reports 19,00% and 2,00% as separate groups.
- `ProviderSpec`: the two issuers are told apart by domicile.
- Full suite green; CI runs it again on PostgreSQL.


taavipertman-glitch:
Superseded by #1937, which carries this commit rebased onto master.

Combined with #1916 because the order matters: without #1916's per-check-type delete scoping, the 25-day backfill that re-reads these relabelled rows would erase the `check_type='PROVIDER'` audit history this split is meant to preserve.

Note the migration is renumbered **V1_252 → V1_255** — #1934 claims V1_252–254, so #1937 must merge after it.

Branch `feature/limit-check-issuer-grouping` is left in place.

PR #1911, opened 2026-08-27
Correct the fund settlement date: T+4, dealt on the domicile, counted on TARGET2

taavipertman-glitch:
Three separate things were wrong with the fund settlement date. The first was found by reading the code; the other two by measuring 222 executions in the transaction registry against what actually settled.

**Scope warning for review: every fund settlement date moves by one business day.** This started as the calendar fix in the title and grew; the day-count change is the one with real blast radius.

## 1. The calendar came from the manager, not the fund

`fundCalendar` derived the domicile from the allocation row's `Provider` — but a `Provider` is a fund **manager**, and a manager runs funds in more than one domicile.

| ISIN | Fund | Manager | `country` | Calendar used |
|---|---|---|---|---|
| `LU0826455353` | iShares Euro Aggregate Bond Index Fund | Blackrock Luxembourg SA | `LU` | 🔴 Ireland |
| `LU0839970364` | iShares Global Government Bond Index Fund | Blackrock Luxembourg SA | `LU` | 🔴 Ireland |

Both carry `provider = 'ISHARES'`, declared `IRELAND`. `instrument_reference.country` already records the real domicile per instrument, so resolution is now:

```
instrument_reference.country  →  provider's domicile (warns)  →  TARGET2 (warns)
```

The provider stays as a fallback because `country` is nullable, and dropping straight to TARGET2 would be worse than the old guess. `Domicile` gains its ISO code and a `forCountryCode` lookup.

## 2. Funds settle T+4, not T+5

`FUND_SETTLEMENT_BUSINESS_DAYS` was 5. The registry's **79 executed fund orders** say otherwise:

| rule | rows explained |
|---|---|
| T+4 | **61 / 79 (77%)** |
| T+5 | 6 / 79 (8%) |

T+4 is the mode for **every one of the seven funds** we hold — 25/33 for `IE00BFG1TM61`, 13/21 for `IE0009FT4LX4`, 10/15 for `IE00BKPTWY98` — so this is not one fund dragging an average. ETFs were already right: 128 of 143 settle exactly T+2.

## 3. The domicile calendar was applied to the wrong step

The calendar was used to count the settlement days themselves. It shouldn't be: **two executions settled on 1 June 2026, an Irish bank holiday.** Counting settlement days on the Irish calendar cannot produce that date, so the model was falsified, not merely imprecise. No settlement ever landed on a TARGET2 holiday.

The two calendars do different jobs, and now do them separately:

| step | calendar |
|---|---|
| which day the order is **dealt/accepted** | the fund's domicile |
| counting the **settlement** days after that | TARGET2, as for ETFs |

## Also: a guard against the fallback going unnoticed

Adding an instrument is SQL, not a deploy, so a row can arrive without a `country` and settle silently on the manager's domicile. `InstrumentDataValidator.checkFundDomicile` now raises a **FAIL** for a fund allocation whose country is missing or outside IE/LU/FR, carried by the hourly `InstrumentValidationJob`. ETFs are exempt — they settle on TARGET2 wherever they are domiciled.

## What this does not fix, and cannot

T+4 on TARGET2 explains ~75–80% of executions. **The rest is not a calendar problem and no holiday map will fix it:** on 8 of 28 order dates, funds with the *same* manager and *same* domicile settled on different days.

```
order 2026-03-10:  IE00BFG1TM61 → 16 Mar    IE0009FT4LX4 → 20 Mar
order 2026-08-24:  IE0009FT4LX4 → 27 Aug    IE0005032192 → 28 Aug
                   IE0031080751 → 28 Aug    IE00BFG1TM61 → 1 Sep
```

That spread is per-fund non-dealing days — the sub-fund's own underlying markets, prospectus clauses (a)/(b) — which BlackRock publishes for its Asia/EM funds but not for these. A UK delegate calendar was tested and rejected: only 7 rows can discriminate it, and the 2 it appears to fix are equally explained by the Irish calendar, since 4 May 2026 is both a UK and an Irish bank holiday.

**So the calculated date is an estimate.** The authoritative date is the one SEB reports, and comparing the two per ISIN is the follow-up worth doing — it would catch the residual without needing anyone's schedule.

## Verification

Full suite: **8387 tests, 1124 classes, 0 failures.**

Four existing tests were asserting the behaviour this PR removes, so they were rewritten rather than re-dated. Three of them would otherwise have gone green while covering nothing: with an order on 12 March, the Irish and TARGET2 calendars give the same answer, so the provider-fallback tests stopped discriminating entirely. They now order *on* St Patrick's Day, where the fallback still changes the result.

The GAS side was aligned separately (`SETTLE_CYCLE_DAYS`, `SETTLE_DAYS_FUND`) so the sheet and the service don't disagree by a day.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
@@ -1,7 +1,29 @@
 package ee.tuleva.onboarding.investment.calendar;
 
+import java.util.Arrays;
+import java.util.Optional;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.jspecify.annotations.Nullable;
+
+@Getter
+@RequiredArgsConstructor
 public enum Domicile {
-  IRELAND,
-  LUXEMBOURG,
-  FRANCE
+  IRELAND("IE"),
+  LUXEMBOURG("LU"),
+  FRANCE("FR");
+
+  private final String countryCode;
+
+  /**
+   * Resolves an ISO 3166-1 alpha-2 country code, as stored in {@code instrument_reference.country}.
+   * Empty for a null, blank or unsupported code — the caller decides what to fall back to.
+   */
+  public static Optional<Domicile> forCountryCode(@Nullable String countryCode) {
+    if (countryCode == null || countryCode.isBlank()) {
+      return Optional.empty();
+    }
+    String normalized = countryCode.strip().toUpperCase();
+    return Arrays.stream(values()).filter(it -> it.countryCode.equals(normalized)).findFirst();
+  }
 }
--- a/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
@@ -15,10 +15,6 @@ public enum Domicile {
 
   private final String countryCode;
 
-  /**
-   * Resolves an ISO 3166-1 alpha-2 country code, as stored in {@code instrument_reference.country}.
-   * Empty for a null, blank or unsupported code — the caller decides what to fall back to.
-   */
   public static Optional<Domicile> forCountryCode(@Nullable String countryCode) {
     if (countryCode == null || countryCode.isBlank()) {
       return Optional.empty();
--- a/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/calendar/Domicile.java
@@ -1,6 +1,7 @@
 package ee.tuleva.onboarding.investment.calendar;
 
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.Optional;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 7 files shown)


taavipertman-glitch:
**Verified against prod, 27.08.2026** — ran the `FUND` country query from the PR body. All seven `FUND` instruments carry a country, so there are no nulls and the provider fallback never fires for any currently held instrument:

| ISIN | Fund manager | country |
|---|---|---|
| `IE0005032192` | BlackRock Asset Management Ireland Ltd | IE |
| `IE0009FT4LX4` | BlackRock Asset Management Ireland Ltd | IE |
| `IE0031080751` | BlackRock Asset Management Ireland Ltd | IE |
| `IE00BFG1TM61` | BlackRock Asset Management Ireland Ltd | IE |
| `IE00BKPTWY98` | BlackRock Asset Management Ireland Ltd | IE |
| `LU0826455353` | Blackrock Luxembourg SA | **LU** |
| `LU0839970364` | Blackrock Luxembourg SA | **LU** |

So after this merges every fund resolves its calendar from `instrument_reference.country`, and `Provider.getDomicile()` is not consulted for any live instrument. The two Luxembourg rows are exactly the ones that were being settled on the Irish calendar via `ISHARES`.

The fallback stays in place for instruments added later without a country recorded — it is dead for today's data, not redundant.

PR #1915, opened 2026-08-27
Give every PostgreSQL test context its own database

vootelerotov:
Every Spring test context that resolves pg or ci now runs against its own database, cloned from a per-worker pre-migrated template inside the shared server. CI's two Gradle worker JVMs share one Postgres service container, so contexts could see each other's committed rows; several tests had been rewritten to assert deltas instead of absolutes purely to tolerate that.

The template name is per worker because both JVMs share one server: a single name would let one worker clone a template the other is still migrating.


vootelerotov:
Merge it in, lets see if there is any performance impact. In worse case, we revert.

PR #1916, opened 2026-08-28
Let each check type own its own limit check event row

taavipertman-glitch:
## The problem

`LimitCheckService.checkFund` deleted **every** event for a fund and date, then rewrote all four check types:

```java
limitCheckEventRepository.deleteByFundAndCheckDate(fund, checkDate);   // all types
saveEvent(fund, checkDate, POSITION,  positionBreaches);
saveEvent(fund, checkDate, PROVIDER,  providerBreaches);
saveEvent(fund, checkDate, RESERVE,   reserveBreach);
saveEvent(fund, checkDate, FREE_CASH, freeCashBreach);
```

Harmless while all four are written. Destructive the moment one stops: `LimitCheckJob.backfillLimitChecks` re-runs `checkFund` over a **25-day window**, so the first backfill after a check type is retired deletes that type's rows across the window and does not put them back.

That matters now because the provider concentration check is going away, and the `check_type = 'PROVIDER'` rows are the audit record it has to leave behind (A2 in the TKF100 constraints register). This PR is the prerequisite that makes the removal safe. **It does not remove anything** — the provider check still runs, as register entry L10 requires until A2 leaves the prospectus.

## The change

The delete now names the check type it is about to rewrite, so a retired type's rows are never touched.

The three `saveEvent` overloads now only decide `breachesFound` and the result map, handing both to one `replaceEvent` — removing three copies of the same builder. `deleteByFundAndCheckDate` is gone; compilation proves it had exactly one caller.

## Why this is a safe refactor

`LimitCheckIntegrationTest.rerunReplacesExistingEventsInsteadOfCreatingDuplicates` is **unchanged** — still `hasSize(4)`, still all four check types — and still green. That is the proof nothing observable changed.

New coverage:
- `LimitCheckEventRepositoryTest.deletingOneCheckTypeLeavesTheOtherCheckTypesAlone` — two types saved, one deleted by type, the other survives **by id**.
- `LimitCheckServiceTest.deletesOnlyTheCheckTypesItRewrites` — one scoped delete per check type, `times(4)` saves.

## Incidental

`checkFund` is not `@Transactional`, so each delete/save pair runs in its own transaction. Previously a crash between the wide delete and the saves could leave the fund and date with **no** events at all; now the worst case is one check type's row. Not the point of the PR, but strictly better. A real atomicity fix is a separate change.

## Test

`./gradlew build` green on **H2** and on **real PostgreSQL 17** — full suite both times, no failures. The PostgreSQL run used `SPRING_PROFILES_ACTIVE=ci,test` against a native server (no Docker needed; `ci` takes Postgres from localhost, only the `pg` profile uses Testcontainers), with the per-context databases introduced in feb13288.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckEventRepository.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckEventRepository.java
@@ -11,5 +11,6 @@ interface LimitCheckEventRepository extends JpaRepository<LimitCheckEvent, Long>
   List<LimitCheckEvent> findByFundAndCheckDate(TulevaFund fund, LocalDate checkDate);
 
   @Transactional
-  void deleteByFundAndCheckDate(TulevaFund fund, LocalDate checkDate);
+  void deleteByFundAndCheckDateAndCheckType(
+      TulevaFund fund, LocalDate checkDate, CheckType checkType);
 }
--- a/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/limit/LimitCheckService.java
@@ -150,7 +150,6 @@ class LimitCheckService {
     var freeCashBreach =
         freeCashLimitChecker.check(fund, cashTotal, liabilityTotal, pendingCashImpact, fundLimit);
 
-    limitCheckEventRepository.deleteByFundAndCheckDate(fund, checkDate);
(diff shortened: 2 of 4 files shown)


taavipertman-glitch:
Superseded by #1937, which carries this commit rebased onto master and merges it ahead of the Xtrackers issuer split from #1907.

Kept together so the sequence cannot be got wrong — this delete scoping has to be in place before the backfill re-reads the relabelled provider rows.

Branch `refactor/scope-limit-check-event-delete` is left in place.

PR #1936, opened 2026-09-01
Apply the fee policy per accrual date, and anchor its back-extension on the fund

### LINKED ISSUE #1856: Apply the fee policy per accrual date on the three NAV-facing readers

### LINKED ISSUE #1887: Fee policy back-extension is anchored on the earliest row, not the fund's inception

taavipertman-glitch:
Supersedes #1897 and #1890, which are closed in favour of this. Closes #1856. Closes #1887 — the issue #1890 was opened against, which would otherwise stay open once this ships its fix.

They are combined because they are not independent: #1890 retargets the
back-extension inside `FeeChargedToFundPolicy.Resolver`, and #1897 makes
`chargedOn` run over *every* accrual date instead of one. Separately they
conflict on the same lines; together, #1890 stops being cosmetic and becomes
load-bearing — asked about a whole month, a series reaching back past the first
policy row has to resolve rather than throw.

## The bug

A month-to-date accrual is a sum over many days, but the NAV-facing readers
asked "is this fee charged to the fund?" **once**, for a single day, and applied
that answer to the whole sum. Tuleva bears TUK75's depot fee; the board moves it
to the fund on 15 September. Asking once for the 30th admits the **entire**
month into NAV; asking once for the 1st admits none of it.

It is still live on master behind the new `NavFees` port: `recordDailyFees`
saves an accrual row for *every* day and only skips the **ledger** entry when
the policy does not charge it, so `getUnsettledAccrual` sums uncharged days too,
and `NavCalculationService.navFacingAccrual` then re-gates the whole month on
one day's answer.

## The fix — one seam, not three

- `FeeAccrualRepository.getAccruedFeesByDateForMonth` / `getUnsettledAccrualByDate`
  return the same rows grouped by `accrual_date` instead of pre-summed.
- `FeeChargedToFundPolicy.Resolver.sumChargedDays(...)` sums only the days the
  policy charges, reusing the resolver that already loads the rows once, so it
  costs no extra queries.
- `FeeCalculationService`, `TransactionInputService.getAccruedFees` and
  `FeeAccrualPositionSyncJob.chargedAccrual` all filter per day.
- `NavCalculationService.navFacingAccrual` is **deleted deliberately** — with the
  policy applied upstream per date, re-gating the summed figure would undo the fix.

Scale is preserved exactly: the filtered sum rounds to 2 like the
`ROUND(SUM(...), 2)` it replaces and returns a plain `0` for no rows.

## Reviewer's attention, please

`NavFees.chargedToFund` had **no production caller left** once `navFacingAccrual`
went, so it is removed from the port, from `NavFeesAdapter` and from its two
adapter tests, per the repo's "no shims, update all callers" rule. Say so if you
would rather keep the port method.

`TransactionInputServiceTest.gatherInput_leavesOutADepotFeeThatIsExcludedFromNav`
no longer asserts the repository is *never queried* — under the new design it is
queried and the policy excludes the days. It now stubs a real depot accrual and
asserts the result is zero, which is the stronger claim.

## Tests

`sumChargedDays_splitsAMonthThePolicyFlipsInsteadOfTakingOneDaysAnswer` is the
case this exists for: policy false to 14.09, true from 15.09, only post-flip days
summed. `sumChargedDays_stillRefusesAGapRatherThanSkippingTheDay` pins that an
uncovered day is an error, not a silent zero — and it is now asked about every
accrual date, so a mid-month gap is caught instead of skipped.

95 tests green across all 8 touched classes locally; CI is the full check.


taavipertman-glitch:
Rebased onto current master (`f03091aa5`) and added a ratchet commit.

The earlier CI failure was `CohesionMetricsTest`: the per-accrual-date split adds one class the LCOM4 metric counts as disconnected. Master has since raised the baseline 451 → 454 in #1919; this PR measures **455**, so the ratchet moves 454 → 455 in its own commit, following the convention #1919 used.

⚠️ **Interacts with #1918**, which independently needs 457. `metrics/baseline.json` is the one file the two conflict on. Merged together the true count is **458** (verified by merging all open PRs and measuring), so whichever of the two lands second must resolve the conflict to 458 rather than keeping its own number.

PR #1937, opened 2026-09-01
Scope the limit check event row per check type, then split Xtrackers into its two issuers

taavipertman-glitch:
Supersedes #1916 and #1907, which are closed in favour of this.

They are combined because the order between them is load-bearing. #1916 scopes
`LimitCheckService`'s event delete per check type; #1907 relabels providers and
its rows are re-read by the 25-day backfill. With the old wide delete still in
place, that backfill would erase the `check_type='PROVIDER'` audit rows — the
very history the issuer split is meant to preserve. One PR makes the sequence
impossible to get wrong.

## 1. Let each check type own its own limit check event row

`LimitCheckService` deleted every event for a fund/date before writing, so a
run of one check type wiped the other's row. Scoped to the check type being
written, so `POSITION` and `PROVIDER` history survive each other.

## 2. Split Xtrackers into its two issuers

The provider limit is measured **per issuer of the units**, not per management
company. `IE00BJZ2DC62` is issued by Xtrackers (IE) plc and `LU0476289540` by
the Luxembourg SICAV — two legal issuers under one manager, DWS Investment S.A.
Checked as one group they are 19,00% + 2,00% = **21,00% against a 20% limit**;
as two they are 19,00% and 2,00%.

`V1_255` relabels only the **version in force**. Earlier versions keep
`XTRACKERS` — they record what was approved and what the check said on those
days — and `Provider` keeps the value so they still read. The limit row follows
the split in the same migration: an issuer with no limit row is dropped from the
check without a word, so the two must move together.

## ⚠️ Merge order — this must land AFTER #1934

The migration was renumbered **V1_252 → V1_255** because #1934 claims V1_252–254.
Flyway is strict-ordered here (`out-of-order` is never enabled), so if this
deploys first, #1934's deploy fails validation and ECS rolls back while CI stays
green. Re-check the number again if anything else merges a migration meanwhile.

## Not removed

`ProviderLimitChecker` stays. Removing it is gated on the provider 20% coming
out of the prospectus, which has no date yet. `ModelPortfolioAllocation.provider`
is load-bearing beyond the limit check — fund-manager display name in the
investment report, plus `SettlementDateCalculator` domicile.

180 tests green across the limit / instrument / portfolio packages locally.


taavipertman-glitch:
Force-pushed: this PR was missing a commit.

When I consolidated #1916 and #1907 I cherry-picked only #1907's tip commit, silently dropping `Stop H2 choking on the instrument reference check constraints` — the `V1_249_1__drop_alter_added_check_constraints.sql` H2 migration and `InstrumentReferenceRulesTest` (originally the work from the closed #1906). Both are restored, and the branch is now an exact line-for-line union of the two source branches.

That gap would not have shown up in a targeted run: H2 only chokes on ALTER-added CHECK constraints once a second Spring context opens the shared in-memory database, so it reproduces in the full suite and never when a class runs alone.

Also rebased onto current master (`f03091aa5`). Cohesion count measures 454 against the new 454 baseline, so no ratchet change is needed here.

PR #1940, opened 2026-09-02
Give the escalation threshold migration a version of its own

taavipertman-glitch:
## master is red — Flyway cannot start

Two migrations claim version **1.252** on master right now:

| file | origin |
| --- | --- |
| `V1_252__savings_fund_onboarding_updated_at.sql` | already on master |
| `V1_252__align_escalation_net_td_threshold_with_daily_threshold.sql` | #1934, merged 07:54:48 |

#1934 merged without renumbering; #1937 merged 57 seconds later and took 1.255.

Flyway rejects the set while scanning:

```
FlywayException: Found more than one migration with version 1.252
```

The application context never starts, so `ci/circleci: build` fails on master and **every `@DataJpaTest` and `@SpringBootTest` in the suite** fails to load with it. Nothing can deploy until this is in.

## The fix

Renumber the newer one to **1.256**, the first free version.

**Nothing had applied it.** Flyway refuses the whole set before running anything, so no environment has run any of #1934's three migrations — the collision arrived in the same commit that introduced them.

**Ordering is not load-bearing.** 1.253, 1.254 and this one are independent `INSERT`s into `investment_parameter` under distinct `parameter_name`s (`ESCALATION_THRESHOLD_DAYS`, `TD_RESIDUAL_TOLERANCE_ANNUAL`, `ESCALATION_NET_TD_THRESHOLD`), so resequencing one past the other two changes no outcome.

## Verified

Ran `FeeChargedToFundPolicyTest` — one of the `@DataJpaTest`s that fails on master — against this branch: **BUILD SUCCESSFUL**, 6/6, zero Flyway duplicate-version errors. The context loads again.


vootelerotov (approved):
LGTM!

PR #1943, opened 2026-09-02
Renumber the duplicate Flyway migration V1_252 to V1_256

vootelerotov:
Two merged branches both took version 1.252: the savings fund onboarding updated_at column (PR #1919, merged 1 September and deployed the same day) and the tracking-difference escalation threshold (PR #1934, merged 2 September). Flyway refuses to start with two migrations at one version, so every Spring context on master has failed since the second merge and nothing can deploy.

Production's flyway_schema_history holds 1.252 as "savings fund onboarding updated at", so that file keeps its version; the escalation migration has never run anywhere and moves to the next free number. It is a standalone investment_parameter insert, so running after 1.253-1.255 changes nothing.


vootelerotov:
Superceded.

PR #1945, opened 2026-09-03
Migrate Smart-ID to RP API v3 with device-link login

ErkoRisthein:
Upgrades `ee.sk.smartid:smart-id-java-client` from 2.3.1 to 3.2 and moves both Smart-ID flows to RP API v3. Supersedes #1505, which was the bump alone.

## Login

Login no longer starts from a personal code, which is what the phone scams exploit: a fraudster could start a session on a victim's code and talk them through PIN1. Every attempt now opens an anonymous device-link session.

- Desktop: the backend rotates a QR link every second from the session secret.
- Phone: a same-device link returns to the frontend callback, whose token and session secret digest are verified before the user challenge verifier is handed to the SDK validator.
- The flow type of the finished session must be one we offered.
- A browser that completes such a login gets a signed, HttpOnly remembered-account cookie. Only that browser may start a push login by document number, with the verification code choice interaction. The cookie is dropped when Smart-ID reports the account gone or the visitor says it is not their account.

Login completion moved into the token request: a single session status call, cached in the HTTP session once complete, replaces the background poller and its write-back.

## Signing

Stays notification-based. When the login stored a document number the signing certificate is fetched silently, otherwise a notification certificate choice runs first. Signatures are requested as `sha256WithRSAEncryption` because DigiDoc4j 6.1.1 cannot finalize RSASSA-PSS.

## Identity restrictions

- Smart-ID accounts from other countries are refused.
- ID-card login requires an Estonian certificate subject country, and Web eID requires the Estonian personal code prefix.
- E-resident digital IDs and EU citizen ID-cards are refused for AML and KYC reasons.

## Configuration

User certificates validate against SK CA certificates bundled per environment; TLS keeps the existing trust store. Scheme name and callback URL are configured per environment, and the demo relying party identity follows SK's v3 values.

## Deploy order

Deploy this before the client, whose login start sends the new `flow` field. Frontend counterpart: TulevaEE/onboarding-client#1634 (branch `smart-id-v3-login`).

## Before production

- Confirm with SK that our live relying-party UUID may call `/v3` anonymous device-link, and whether callback domains need registering.
- The production truststore pins SK TLS certificates that expire 2026-10-10.
- No rate limiting yet on the anonymous start endpoint.


ErkoRisthein:
## Security follow-ups

Measured against SK's [additional security measures](https://sk-eid.github.io/smart-id-documentation/rp-api/additional_security_measures.html). Three of their seven are already satisfied by this PR: anonymous device links, users can download their signed documents, and session fixation is handled (`SecurityConfiguration.java:89` starts a new session on login). The rest, in the order I would do them.

### Do now, no external dependency

**1. Rate limit the anonymous device-link start.** `POST /v1/smart-id/login` is unauthenticated and creates a session at SK on every call, so any visitor can burn sessions from our IP. SK authenticates relying parties by IP plus UUID, which makes that a shared-fate problem across all our logins. Per-IP and per-browser-session throttle, plus a cap on concurrent pending sessions per session.

**2. Record the IP and browser on login events.** `LoginEventBroadcaster` logs method and document type only. SK's monitoring measure wants connection data next to it, and without the IP we cannot see an attack in progress or answer questions about one afterwards.

**3. Shorten the remembered-account cookie.** `RememberedSmartIdAccounts` defaults to 365 days. A year is long for a cookie that decides who may be pushed a login. Suggest 90 days, and re-verify with a device link after that rather than extending on each use.

**4. Let people forget every remembered browser.** "Not you?" clears the cookie on the browser in front of you. SK asks that users be able to invalidate all of them, which means a per-browser identifier stored server side and an action in account settings.

### Needs SK first

**5. Request the `shareMdClientIpAddress` privilege.** The single highest-value item. It returns the phone's IP in the session status, so a mismatch against the browser IP is a strong signal of a relayed or screen-shared QR, which is the one residual risk of the cross-device flow and the objection LHV raised publicly. Note RP API 3.2.2 throws immediately if the field is requested without the privilege, so the code has to be switchable per environment. Needs a GDPR basis. Start by logging and alerting on mismatches, not blocking.

### Compliance, before production

**6. Cookie consent and privacy policy** for `SMART_ID_REMEMBERED_ACCOUNT`. It carries a document number, so it is personal data, and SK requires the declaration explicitly. This is a release blocker rather than an enhancement.

### Done in the client PR

Access and refresh tokens were being written to the browser console on every successful Smart-ID poll, on master as well. Fixed in TulevaEE/onboarding-client#1634.

Issue #1850, opened 2026-08-12
Follow-ups from the four-PR review round (#1827 #1828 #1839 #1840)


JordanValdma:
Open questions deliberately deferred so the 40 verified fixes could ship. Each needs a human answer, not more code reading.

## 1. SRI trading-day constant is unverified — compliance-relevant

`SriCalculator` now measures VaR over a fixed 5-year holding period instead of the sample size (the actual bug fix). That required two constants:

- `TRADING_DAYS_PER_YEAR = 260`
- `MINIMUM_OBSERVATIONS = 1000`

Both were derived from **this codebase's own conventions** — 52 × 5 weekdays, matching the weekday-only price filter and SRRI's `√52`; and SRRI's 200 weekly observations restated in trading days. **Neither is cited to the RTS**, deliberately: inventing a regulatory citation would be worse than leaving it open.

`250` / `252` / `256` are all plausible and shift VEV by roughly 2%, which can move a published class near a bucket edge.

**Action:** whoever owns risk checks `260` against the ESAs' worked example for Annex II. Grep confirmed the repo has no prior annualisation precedent for risk — the only `DAYS_PER_YEAR` is `365.25` in `WorldMarketBenchmarkService`, a different concept.

## 2. Settlement cycle assumption in the pending-order model

`PendingOrderImpactService` now decides unreported positions by position report date rather than `OrderStatus` — which implements the intent stated in `44941a68f` ("SENT orders **not yet in the SEB report**").

The remaining open question is whether the sell cap is safe: the argument for closing the "can oversell holdings" finding is that a buy settles no later than the sell it funds. That holds for same-cycle settlement.

**Action:** confirm a sell can never settle before the buy funding it. If it can, reopen the sell-cap question.

Note `investment_transaction_execution` currently has **zero rows in 90 days** and `scheduled_settlement_date` is universally NULL, so this is latent rather than active. It activates the first time SEB populates settlement dates.

## 3. Quantity control has never run against real trade data

Same root: zero executions in 90 days while positions moved across 153 nav dates. The `QUANTITY_CHANGE` check therefore computes a budget of 0 for every ISIN. It is quiet only because quantities are not moving.

**Action:** exercise it deliberately before relying on it — the first real trading day is otherwise also its first test.

## 4. Ratify deleting `AccrualRoundingDriftChecker`

[...]

PR #1958, opened 2026-09-08
Block the position import per fund, not per file

taavipertman-glitch:
> ⚠️ **Draft, parked for a future session — do not merge as-is.** The branch is ~400 commits behind `master` and conflicts. It needs a **fresh port**, not a rebase: the packages moved (`ee.tuleva.onboarding.tulevafund.TulevaFund`, `ee.tuleva.onboarding.pipeline.PipelineTracker`) and `TransactionInputService.getAccruedFees` was refactored to `resolverFor(...).sumChargedDays(...)`, so the fee-policy patch below no longer applies as written. Opened to store the work and its reasoning where it can be found.

Three unrelated fixes found reviewing #1827 / #1839 / #1847. **They should be split into three PRs when this is picked up** — they are together only because they came out of one review.

## 1. A negative SECURITY quantity goes back to `FAIL`, because the gate now blocks per fund

#1827 raised it as a `FAIL`, then downgraded it to `WARNING` for a sound reason: `FundPositionImportJob` abandoned the **whole file** for **every fund** the moment any check failed, and the SEB parser maps anything it cannot classify as cash, a receivable, a payable or a register line to `SECURITY` — so a securities borrow or an unmapped custody line would have gated the NAV pipeline on a heuristic.

The fix is the gate, not the severity. `importForProviderAndDate` now drops the positions of every fund whose `HealthCheckResult` has fails and imports the rest normally; the ledger recording and NAV trigger follow the same set. The checks were already scoped to one fund — only the gate was not. With the blast radius down to one fund for one day, a holding you cannot own stops at the gate instead of reaching the NAV input with a Slack line.

⚠️ **#1944 rewrites this same method** and keeps the whole-file blocking, while introducing `changedRowsByFund()`. This sits naturally on top of it and will conflict textually.

## 2. The TD gate says when the check could not run

`NavTrackingDifferenceGate`'s generic `catch (Exception)` only marked the step failed and logged, while its `IncompletePriceDataException` sibling calls `notifyCheckCouldNotRun`. So an error inside the check — a fee-policy gap, for instance — switched the tracking check off for the day in silence.

## 3. An unresolved fee policy warns instead of blocking order creation

`FeeChargedToFundPolicy` throws on a missing, gapped or overlapping `investment_fee_policy` row, and `TransactionInputService.gatherInput` had no catch — so a hand-maintained reference row could stop an operator creating orders for that fund.

It now reserves the accrual as if it were charged (the conservative direction, and small next to the portfolio) and raises `CalculationWarningType.FEE_POLICY_UNRESOLVED`, which reaches the operator on the DRAFT snapshot. `TradeCalculationEngine.warnings()` returned `List.of()` for every non-`REBALANCE` mode and would have swallowed it, so it now carries input-stage warnings in all modes.

## Also

Two test names in `QuantityChangeCheckerTest` lose their settlement wording — the behaviour they pin is right, but the names said the position moves at settlement, which is the premise #1851 corrected.

## Withdrawn from this branch

An earlier version had the quantity control skip an ISIN whose executions carried no unit count, on the claim that *"SEB leaves Quantity empty on a FUND BUY"*. **That claim is false** — 27 transaction rows across the real 2026-02 reports all carry a Quantity, the mutual-fund buy `IE00BFG1TM61` included. What is true is that we **order** a fund buy in amount, so `orderQuantity` is null until the dealing NAV is struck; the execution still carries units. The code was removed: it opened a real blind spot to solve a problem that does not exist.

## Testing

Targeted tests green after the withdrawal (`QuantityChangeChecker`, `TradedQuantitySource`, `CompletenessChecker`, `FundPositionImportJob`, `TransactionInputService`, `NavTrackingDifferenceGate`). The full suite was green on the pre-withdrawal state, against the old base.

Context: tuleva PR #473 §4.15. Related: #1850, #1944.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/check/health/CompletenessChecker.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/health/CompletenessChecker.java
@@ -1,5 +1,6 @@
 package ee.tuleva.onboarding.investment.check.health;
 
+import static ee.tuleva.onboarding.investment.check.health.HealthCheckSeverity.FAIL;
 import static ee.tuleva.onboarding.investment.check.health.HealthCheckSeverity.WARNING;
 import static ee.tuleva.onboarding.investment.check.health.HealthCheckType.COMPLETENESS;
 import static ee.tuleva.onboarding.investment.position.AccountType.CASH;
@@ -38,6 +39,8 @@ class CompletenessChecker {
               "%s: no CASH position found for navDate=%s".formatted(fund, navDate)));
     }
 
+    // A holding you cannot own must not reach the NAV input. This FAILs, and since the import gate
+    // blocks per fund, it takes down only this fund's positions for the day.
     positions.stream()
         .filter(position -> position.getAccountType() == SECURITY)
         .filter(position -> position.getQuantity() != null && position.getQuantity().signum() < 0)
@@ -47,7 +50,7 @@ class CompletenessChecker {
                     new HealthCheckFinding(
                         fund,
                         COMPLETENESS,
-                        WARNING,
+                        FAIL,
                         "%s: negative SECURITY quantity %s for %s"
                             .formatted(
                                 fund,
--- a/src/main/java/ee/tuleva/onboarding/investment/check/tracking/NavTrackingDifferenceGate.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/check/tracking/NavTrackingDifferenceGate.java
@@ -68,6 +68,7 @@ public class NavTrackingDifferenceGate {
       log.warn("TD gate incomplete price data, proceeding: fund={}, date={}", fund, navDate);
       return Optional.empty();
     } catch (Exception e) {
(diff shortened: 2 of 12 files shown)


taavipertman-glitch:
Superseded. The three fixes in here were ported fresh onto master and split into one PR each, as this description proposed:

- **Per-fund import gate** (+ negative SECURITY quantity back to `FAIL`) → #1986
- **TD gate says when the check could not run** → #1984
- **Unresolved fee policy warns instead of blocking order creation** → #1985

Two notes from the port:

- The `QuantityChangeCheckerTest` renames listed under *Also* are already on master — those tests say "Executed" now, so nothing carried over.
- The fee-policy patch needed reworking, not just rebasing. `resolverFor(...).sumChargedDays(...)` moved the throw: `resolverFor` throws when no row exists at all, while `chargedOn` throws per day on a gap or overlap. #1985 catches both; the version here caught only the first.

The claimed textual conflict with #1959 over `FundPositionImportJob` did not materialise — that PR changes the `parser.parse` block and this one the health-check block below it. Closing in favour of the three replacements.

PR #1959, opened 2026-09-08
Refuse a SEB report that carries no As of date

taavipertman-glitch:
> ⚠️ **Draft, parked for a future session — do not merge as-is.** The branch is ~400 commits behind `master` and conflicts; it needs a **fresh port**, not a rebase (packages moved, and `FundPositionImportJob` is rewritten by #1944). It also overlaps #1958, which touches the same method. Opened to store the work and its reasoning.

## The bug

Both SEB parsers fell back to `report_date` when the `As of:` preamble was missing:

- `SebPendingTransactionReconciliationService.asOfDate()` → stamps `investment_transaction_execution.reported_date`
- `SebFundPositionParser` → sets `fund_position.nav_date`, the NAV date itself

`report_date` is the **S3 filename date**, i.e. the day SEB *sent* the file — one business day after the rows it carries. `SebReportHeadersTest` pins the gap: `reportDate(SENT)` = 26 Jan while `AS_OF` = 25 Jan.

The two reports run on **one clock**: a trade appears in the pending-transactions report *and* in the position report's quantity on execution, with the cash held as a liability (buy) or receivable (sale) until settlement. Settlement moves the cash and clears that leg; it never moves the quantity again. So the substitution is not conservative — it double counts. `reported_date` lands a day ahead of `positionDate`, `PendingOrderImpactService.isMissingFromPositionReport` returns true for a fill the position report already carries, and the position is synthesized on top of itself while its cash is reserved once.

**The invariant does not catch it.** Gross and `securityValue` both move by the fill amount, so `netInvestable − securityValue == freeCash` still holds and W1 stays silent. What breaks is the **denominator** — on a €1.9m fill against a €12m fund the bought instrument reads ~9pp overweight and everything else ~7pp underweight, so a REBALANCE proposes selling back what was just bought, and a position limit can read as breached when the true weight is inside it. The sell side mirrors it: `adjustPosition(... .max(ZERO))` can take a sold position to zero and the next rebalance buys it back.

## The fix

A report with no `As of` marker has no honest date, and the filename is a different clock — so it is refused, not dated. The reconciliation returns before matching anything; the position parser throws `MissingReportAsOfDateException`; both publish `MissingReportAsOfDateEvent` → one Slack line on `Channel.INVESTMENT` saying the report was **not used** and that a new report must be requested from SEB and **reviewed on their side**, since a missing preamble means the rest of the file may be wrong too. The 7-day reconciliation lookback and 14-day position lookback pick up the corrected file automatically. An empty report still returns empty — no rows, nothing to misdate.

`FundPositionImportJob` catches the exception itself because `runImport` swallows every exception into a log line; without that catch the report would be skipped silently every day the file stayed broken.

Accepted consequence: while the report is refused the order stays `SENT` with no fill, so the unfilled remainder is still synthesized. Unavoidable with no fill recorded, and it errs toward not re-buying.

## What the fix flushed out is worth more than the fix

Removing the fallback broke 18 tests. The importer strips the **first five rows** of every SEB CSV into `metadata` via `SebReportSource.extractCsvMetadata`, so the preamble is never in `raw_data` — every fixture handing a parser raw rows with `metadata = Map.of()` had **no date at all** and was silently riding the fallback.

Including `NavPipelineIntegrationTest`, which loads a **real** SEB CSV carrying `As of:;2026-02-13` and then called `saveReport(…, Map.of())`, throwing the preamble away. It now goes through `extractCsvMetadata`, i.e. through the production path. That check also confirms the change is safe in production, where the As-of comes from metadata.

Context: tuleva PR #473 §4.16 and Formats Reference §B2.1. Related: #1827, #1851, #1896, #1944.

### DIFF
--- a/src/main/java/ee/tuleva/onboarding/investment/position/FundPositionImportJob.java
+++ b/src/main/java/ee/tuleva/onboarding/investment/position/FundPositionImportJob.java
@@ -20,6 +20,8 @@ import ee.tuleva.onboarding.investment.position.parser.SebFundPositionParser;
 import ee.tuleva.onboarding.investment.position.parser.SwedbankFundPositionParser;
 import ee.tuleva.onboarding.investment.report.InvestmentReport;
 import ee.tuleva.onboarding.investment.report.InvestmentReportService;
+import ee.tuleva.onboarding.investment.report.MissingReportAsOfDateEvent;
+import ee.tuleva.onboarding.investment.report.MissingReportAsOfDateException;
 import ee.tuleva.onboarding.investment.report.ReportProvider;
 import java.time.Clock;
 import java.time.LocalDate;
@@ -151,11 +153,20 @@ public class FundPositionImportJob {
     }
 
     InvestmentReport investmentReport = report.get();
-    List<FundPosition> positions =
-        parser.parse(
-            investmentReport.getRawData(),
-            investmentReport.getReportDate(),
-            investmentReport.getMetadata());
+    List<FundPosition> positions;
+    try {
+      positions =
+          parser.parse(
+              investmentReport.getRawData(),
+              investmentReport.getReportDate(),
+              investmentReport.getMetadata());
+    } catch (MissingReportAsOfDateException e) {
+      // runImport swallows every exception into a log line, so this has to alert here or the
+      // report is silently skipped day after day.
+      log.error("Positions report refused: provider={}, date={}", provider, date, e);
+      eventPublisher.publishEvent(new MissingReportAsOfDateEvent(provider, POSITIONS, date));
+      return new ImportResult(0, 0);
+    }
     log.info(
         "Parsed fund positions: provider={}, date={}, count={}", provider, date, positions.size());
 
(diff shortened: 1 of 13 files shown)


taavipertman-glitch:
Superseded by #1987, ported fresh onto master as this description proposed.

The reasoning here carried over unchanged — the refusal, the one-clock argument, the Slack line on `Channel.INVESTMENT`, and the catch in `FundPositionImportJob` because `runImport` swallows exceptions into a log line.

What changed in the port:

- `NavPipelineIntegrationTest` threw the preamble away in **two** helpers on current master, not one — the reissued-report helper does it too. Both now go through `sebReportSource.extractCsvMetadata`.
- The fixture count is 29 tests across 6 files rather than 18, since the reconciliation ITs grew since this branch was cut. All were fixed by giving them a real As-of date one business day before their report date; no check was softened.

The overlap with #1958 flagged here did not materialise — that PR changes the health-check block, this one the `parser.parse` block above it, so they merge independently. Closing in favour of #1987.

PR #1988, opened 2026-09-14
Accept third-party deposits for unit holders under representation

stenehrlich-tuleva:
**What:** a savings fund deposit naming a unit holder who is under representation — a minor, or an adult under guardianship — is now attributed to them whoever sent it, instead of being returned unless the remitter was the unit holder or their legal guardian. The payment records whether the remitter was a third party, for AML scoring.
**Why:** from 18.09.2026 the fund rules deem a purchase order given by a deposit made by the unit holder "or by a third party for the benefit of the unit holder", so grandparents and friends can gift into a child's account.
**Review:** @vootelerotov

---

**Scope is deliberately narrower than the rules text.** The fund rules say "a third party for the benefit of the unit holder" with no age limit, so a gift to an ordinary adult is a valid purchase order under the 18.09.2026 wording — but this PR still returns it. Tuleva's decision: open it only for the case it is meant to serve. Company accounts are unchanged too.

`hasRestrictedLegalCapacity` reads the `parent_child_link` table the guardian carve-out already uses: minority by age, guardianship by an unexpired `GUARDIAN` link (`registerGuardian` does not require a minor, so adult wards are representable). Suspension is not consulted — a suspended link is an issue with the guardian, it does not give the ward their capacity back.

**Two things worth a look:**

1. **`third_party_deposit` is three-valued, not a boolean.** A remitter name that does not match the unit holder is *not* evidence of a third party: Wise puts its own name on the transfer and Montonio often sends no sender details at all. The name only rules a third party out, never in; unknown stays null. database-administration #15 leaves nulls out of AML scoring, same as it already does for a missing remitter identity.

2. **A mistyped personal code now has fewer guards.** If somebody types a code that passes the checksum *and* belongs to an onboarded represented unit holder, the money lands on the wrong account — the remitter comparison used to catch that. Low probability, not zero. A gift link would remove the hand-typing entirely; not built yet.

Both payment integration tests are byte-identical to master again, which is the point: outside the represented case, nothing about verification changed.

database-administration #15 needs the `third_party_deposit` column this PR adds, so **merge and deploy this first**.


vootelerotov:
Need to re-check migration version before merging.

PR #1996, opened 2026-09-15
Renumber the unit transfer migrations above master

JordanValdma:
## master cannot boot

Two pairs of migrations share a version on `master`:

| version | files |
|---|---|
| **V1_267** | `add_redemption_hold_reason.sql`, `add_unit_transfer_transaction_type.sql` |
| **V1_268** | `add_nudge_to_email.sql`, `create_savings_fund_unit_transfer.sql` |

Flyway refuses to start on a duplicate version:

```
FlywayException: Found more than one migration with version 1.267
```

So `entityManagerFactory` fails, the Spring context never loads, and every repository test fails. Master's last two builds are red for this reason and not for anything in the changes they carried.

## Which pair moves, and why that one

Landing times on master:

| file | landed |
|---|---|
| `V1_267__add_redemption_hold_reason` | 14:41 |
| `V1_268__add_nudge_to_email` | 14:56 |
| `V1_267__add_unit_transfer_transaction_type` | **15:08** |
| `V1_268__create_savings_fund_unit_transfer` | **15:08** |

The unit transfer pair landed last and is what broke the sequence. Nothing has booted since 15:08, so that pair is the only one **certain** never to have been applied in any environment — renumbering it cannot strand an already-applied version, which renumbering 267/268 might.

They move to **V1_269** and **V1_270**, keeping their relative order: `ALTER TYPE ledger.transaction_type ADD VALUE 'UNIT_TRANSFER'` before the table that uses it.

Contents are unchanged; these are pure renames.

## Verification

`UnitTransferRepositoryTest`, `TrackingDifferenceEventRepositoryTest` and `InvestmentAccountRepositoryTest` — three of the suites failing on master — all pass on this branch, i.e. the context loads again.

## How this happened

Two PRs each renumbered "above master" at the same time and collided. This is the incident CLAUDE.md already records for V1_198–V1_202: the renumber has to be re-checked immediately before merge, because a racing PR may have claimed the numbers since.


ErkoRisthein:
Verified on latest master (ecdd77294): both duplicate pairs are still there, and the production Flyway history has 1.267 `add redemption hold reason` and 1.268 `add nudge to email` applied (deployed from c52d616d7 at 15:14), with no unit transfer table present. So moving the unit transfer pair is the only safe direction, and 269 and 270 are still free: nothing that touches migrations has landed on master since your branch point.

On this branch the context boots on H2 and on PostgreSQL through Testcontainers: `UnitTransferRepositoryTest` (9 tests) and `WithdrawalsControllerIntegrationTest` (3 tests) pass. The diff is two pure renames.

For the record, this is the second collision today. The first one, nudge versus redemption hold at 1.267, was fixed by c52d616d7 (nudge moved to 1.268); the unit transfer pair (f82e20a2c, merged 15:08) landed in between and reintroduced two duplicates. Each PR passes CI on its own branch and only collides after merge, so the cheap guard is the branch protection setting "require branches to be up to date before merging" on master: it forces the rebase, and Flyway then fails the PR build instead of master.

One dependency: the payment rate season PR I am about to open takes V1_271 for its new table, so this one should merge first.

PR #1997, opened 2026-09-15
Let the internal ops token call the savings fund unit-transfer endpoints

vootelerotov:
The five unit-transfer handlers took only the admin API token. Their caller is agent-gateway, which carries its credential as an encrypted file in its image, and the admin token reaches every /admin endpoint there is. The ops token is the one meant for internal operational callers, and every other savings fund, ledger, party and KYB admin endpoint already accepts it, so these five now do too, through validateWithOpsAccess.

This reverses what fc9bf759e argued: "the ops token lets someone retry a payout, not give away another person's units." That line was wrong on its own terms. retry-payout is admin-token only, while approve-review on a redemption, which takes a free-text approvedBy exactly like these handlers, has accepted the ops token all along. The two-person rule is unchanged: submittedBy and approvedBy are still compared in the service and by the database constraint, whichever token the caller holds.

The controller test that pinned the ops token out becomes a positive control, and the happy-path tests for all five handlers now run with both accepted tokens, so a handler that stops taking the ops token fails the suite rather than agent-gateway.


stenehrlich-tuleva (commented):
I checked the claims against the code. The change is defensible and probably a security improvement, but the improvement does not land with this PR alone, and one sentence in the description is not accurate.

### The win is conditional, and the condition is not here

`validateWithOpsAccess` accepts **both** tokens, so this does not swap the admin token out, it adds the ops token alongside it. The argument for the change is good: agent-gateway has to hold the admin token today, and the admin token opens every `/admin` endpoint there is, including `retry-payout` and ledger `/adjustments`. Moving it to the ops token shrinks that blast radius.

But nothing here makes agent-gateway drop the admin token. If it keeps using it, the net effect is that the ops token gained unit-transfer power and nothing got safer. Is the gateway-side change queued? That follow-up is what turns this from a widening into a narrowing.

### "every other savings fund, ledger, party and KYB admin endpoint already accepts it"

This one is not true. `retry-payout` in `SavingsFundAdminController` and `/adjustments` in `LedgerAdminController` are both on plain `validate()`. There is a deliberate line in those controllers and the description erases it, which invites exactly the objection the PR is trying to pre-empt.

The interesting part is that the line supports your conclusion anyway, just for a different reason. Looking at what is on each side:

- `/adjustments` (admin only) takes a free-form debit account, credit account, arbitrary parties and an arbitrary amount. Unbounded double entry.
- `reclassifications` and `blackrock-adjustment` (ops) are shape-constrained: correct a specific prior transaction, or one fund on one date.

Unit transfer is firmly in the constrained camp. `preview` refuses if the recipient cannot hold units, the plan hash has to match, `approve` re-runs `preview` and refuses if the ledger would no longer make the transfer, and `existsByPlanHashAndState(EXECUTED)` blocks double execution. So the principle is *arbitrary vs shape-checked*, not *admin vs ops by endpoint count*, and on that principle unit transfer belongs next to `reclassifications`. Worth putting that in the description instead, it is a much stronger argument.

### `preview` becomes a balance oracle for the ops token

Not mentioned in the description. `preview` returns `Planned`, which carries `giverUnitsOwned` and `giverPaidIn`, and the controller returns it unfiltered. Anyone holding the ops token can now ask how many units an arbitrary party holds and how much they paid in. Given the ops token already reaches ledger reclassifications and the IBAN whitelist this is probably fine, but it should be a deliberate yes rather than a side effect nobody noticed.

### On the two-person rule

Agreed that it is unchanged, and the service check plus the DB constraint are both there. One precision point, since the description leans on it as reassurance: the caller supplies **both** `submittedBy` and `approvedBy` as free text, so a single holder of either token satisfies the rule by sending two different strings. It is a control against operator error, not against a compromised credential. True before and after this PR, but it means the rule should not carry much weight in the argument about which token to accept.

Minor, pre-existing: the service compares with `strip().equalsIgnoreCase()` while the DB constraint is `approved_by <> submitted_by`, which is case sensitive. The DB is the weaker of the two, so it is a backstop rather than an equal guarantee.

### Tests

Good. The negative controls survive (`withoutAnAdminTokenNothingIsReachable`, `anApprovalWithAnUnacceptedTokenMovesNothing`, `submittingWithAnUnacceptedTokenRecordsNothing`), and turning the pin-out test into a parameterised happy path over both tokens is the right move: a handler that stops accepting the ops token now fails the suite instead of failing agent-gateway in production.

Not blocking. Mainly I would like to know the agent-gateway side is coming, and I would fix the "every other endpoint" sentence.

PR #2005, opened 2026-09-16
Match a guardian by name when the bank sends no payer personal code

stenehrlich-tuleva:
**What:** a savings fund deposit naming a child in the description is now attributed to the child when the remitter's name matches one of the child's registered representatives, instead of being returned because the name did not match the child's own.
**Why:** no bank is obliged to send the payer's personal code — Citadele omits it on 4 of our 5 payments from them — and without it the existing guardian carve-out never runs, so a parent's deposit for their child bounced.
**Review:** @vootelerotov

---

**Overlaps #1988.** That PR accepts a deposit for a represented unit holder from *any* remitter, which subsumes this case. This one is the narrow bug fix: no migration, no `database-administration` dependency, so it can land first. Do not land both blind — whichever merges second needs a rebase in the same lines of `process()`.

**Two things worth a look:**

1. **Statuses.** The name path uses the same ACTIVE + PENDING_KYC widening `isAuthorizedRemitter` already grants the code path. Deliberate: the weaker evidence must not become the more permissive door.

2. **Booking date on the existing path.** A Codex review flagged that the name path asking as of the booking date while `isAuthorizedRemitter` still asked as of today made the name-only path laxer than the code path. So `findRepresentation` gained an `asOf` overload and both attribution paths now ask as of the day the money arrived. `isActiveRepresentation` is authorization, not attribution — untouched, still as of today.

**Not fixed here:** `findPayer` still identifies nobody when the bank sends no payer code, so a payment of this shape that genuinely fails notifies no one.


stenehrlich-tuleva:
Superseded by #1988, which accepts a deposit for a represented unit holder from any remitter and so covers this case already. Closing rather than carrying two PRs over the same lines of `process()`.

PR #2008, opened 2026-09-17
Let a parent hand out a gift link, and a stranger pay into it

stenehrlich-tuleva:
**Mis:** A parent hands out one gift link per child, and whoever holds it pays into that child's savings fund account without logging in.

**Miks:** It removes the two ways a gift used to go wrong — a mistyped but valid personal code sending money to a stranger's child, and a payment from a bank that never passes on the payer's identity, which left it outside the AML rules.

**Review:** @vootelerotov

---

Client side: TulevaEE/onboarding-client#1649. This one merges first.

Worth knowing, because none of it is visible from the diff:

- **The link enforces nothing about eligibility.** Whether a third-party deposit is accepted is already decided when the money arrives, by the rule that governs every other one, so a child who has since turned 18 has the gift returned rather than the link quietly going dead. One gate, not two.

- **A gift is told apart from the parent's own deposit** by having been started through the link, or by coming from somebody who does not act for this child. Nothing on the payment itself separates them: both are third-party deposits. That is why every link payment is recorded, not only the ones carrying a greeting. Known gap: a gift transferred by hand from a bank that sends no payer id code cannot be attributed, and is left out rather than guessed at.

- **The payer is genuinely optional now** rather than faked, so `PaymentReference.personalCode` and `MontonioOrder.billingAddress` are nullable. That is what surfaced the third pillar callback quietly assuming a payer exists. It does, because only the savings fund has anonymous payers, but the assumption was unwritten and is now a `requireNonNull` that says so.

- **A greeting is filed against the payment description**, the only thing that comes back whichever way the money reaches us. Two gifts in the same second share one, so both are shown with no message rather than one of them with the wrong words.

- **`GiftLinkServiceDatabaseTest` runs against a real database on purpose.** Replacing a link passed against mocks and failed against the constraint: Hibernate runs inserts before updates, so the replacement claimed `open_for_recipient` while the old row still held it and the parent got a 500 every time.

- **The cohesion ratchet moved 460 → 464**, which you asked for in the comment below.

**Still open:** `POST /v1/gift-links/{token}/payments` is unauthenticated and writes — it creates a Montonio order and a row. The token is 128 bits so it cannot be found by polling, but whoever holds a real link can call it as often as they like, and there is no rate limiting anywhere in this repo to hang it off. First one here, or separate work?

**Before merging:** the gift link page is the first Tuleva page a non-saver sees, and it shows the child's name and personal code. The personal code is deliberate — without it there is no way to pay from a bank Montonio does not cover, or above its ceiling. Compliance has not signed that off.


stenehrlich-tuleva:
@vootelerotov a question about the cohesion ratchet before this goes further.

This branch sits at 464 disconnected classes against the committed baseline of 460, and that is the only red check: 9058 tests run, one fails. Master is green, so the four are genuinely additive.

They are the two new JPA entities, `GiftLink` and `GiftMessage`, where a Lombok getter per field makes each field its own component, and the two controllers, `GiftLinkController` and `PublicGiftLinkController`, where different endpoints use different collaborators. Lombok builders and records are not affected — `build()` reads every field, and record accessors are already excluded as boilerplate.

I could not find a way to get under 460 that was not worse than the number: adding a method that touches every field to make an entity look cohesive, or splitting a controller per endpoint. So the question is yours to answer rather than mine:

1. Raise the baseline to 464, naming the four classes in the commit?
2. Offset by making four existing classes cohesive, which would be unrelated to this feature?
3. Or is this a sign the metric should treat a Lombok getter on an `@Entity` the way it already treats a record accessor — as boilerplate rather than behaviour? That would move the baseline for everyone, so it is a bigger call.

The file has only ever been lowered so far, both times by you, which is why I did not want to raise it unilaterally.

vootelerotov:
Raise the baseline here :)

vootelerotov (approved):
LGTM!

PR #2010, opened 2026-09-17
Add rules for checks and alerts, and enforce migration ordering

taavipertman-glitch:
Four coding rules for `AGENTS.md`, and one check that enforces a rule the file has carried since V1_198.

## Checks and alerts (new section)

A check that cannot fire is worse than no check: the digest reports it as verified, so nobody looks again. Two failure shapes have both shipped:

- a filter on a column the writing path never populates matches nothing, so the check's input stays `0` while every test passes;
- a comparison between two aggregations over the same column is bounded by the rounding they differ in, so it can never exceed its own tolerance.

Both are invisible to a unit test that stubs the repository — such a test reaches the branch only by supplying a value the query cannot produce, which proves the branch exists rather than that anything reaches it. So: every check reading the database gets an `*IT` over real rows.

The section also asks for what *clears* a finding to be named alongside what fires it. A check that runs for a fund it can never observe returns the same non-answer every period and can never clear — a static fact about the system, not a periodic finding. Ops noise is a defect: operators learn to skim a message that is mostly boilerplate, and the period it carries a real FAIL, nobody reads it.

## Rationale lives in identifiers

The no-comments rule already existed but kept being read as licence to delete the reasoning with the comment. It moves into the name instead — `SISEKORD_4_P_11_7_FIRST_ESCALATING_BREACH_DAY` rather than a comment citing the rule above `ESCALATION_THRESHOLD_FALLBACK`. Applies to SQL and migrations; test names carry the narrative.

## No unused parameters

With the `@EventListener(X.class)` no-arg form named explicitly, since that is the shape most often missed, and the framework-mandated exceptions listed (`RowMapper`'s `rowNum`, SpEL-referenced params, interface implementations) so the rule is not applied where it cannot be.

## `.githooks/migration-order`

Rejects a version at or below master's max, a version claimed twice, and a gap above master's max. Runs at `pre-push` and as a CircleCI job — **CI is the load-bearing half here**, the mirror image of `pii-scan`: the numbering race happens *after* the push, when a PR that merges first takes the numbers, so only a re-run against the new master can see it. H2 compat migrations compare on the full version, so `V1_270_1` beside master's `V1_270` is accepted.

`.githooks/migration-order-test` pins 9 cases in both directions — a checker that cries wolf gets bypassed, which is worse than no checker. Verified against this repo with synthetic clashes: all three failure modes report correctly against the live master max.

## Deliberately not done

PMD's `UnusedFormalParameter` was the obvious enforcement for the unused-parameter rule. Measured it: 33 violations, ~11 of them framework-mandated and unfixable. A third false positives in a ratcheted metric is the same cries-wolf failure, so the rule stays prose and the code-quality bot stays the enforcement. The ~22 genuinely-dead parameters are tracked in #2011.

No production code touched.


JordanValdma (approved):

AGENTS.md
```diff
@@ -133,11 +134,21 @@ Prefer: `text` over `varchar` unless the length is a domain invariant (e.g., `va
...
+- **Ops noise is a defect, not a cosmetic issue.** Operators learn to skim a message that is mostly boilerplate, and the month it finally carries a real FAIL, nobody reads it.
+- **One entity's failure must not take the others down**, and a check that could not run says so rather than returning nothing — an empty result reads as "nothing wrong".
+
 ## Code Style
 
 ### Project-Specific Conventions
 - **Log/exception format**: `"Description: param1=value1, param2=value2"` — greppable
-- **No comments, no Javadoc** on implementation classes. Extract well-named methods instead
+- **No comments, no Javadoc** on implementation classes. Extract well-named methods instead. This binds hardest on *rationale* — the regulation behind a threshold, why a window opens where it does, why a branch is safe. The rationale is not deleted, it moves into the identifier: `SISEKORD_4_P_11_7_FIRST_ESCALATING_BREACH_DAY`, not a comment citing the rule above `ESCALATION_THRESHOLD_FALLBACK`. Applies to SQL and migrations too. Test names carry the narrative; a test needing a comment to say what it pins is misnamed
```
ErkoRisthein, line 151: Use English variable/constant names

taavipertman-glitch: Pushing back on this one, though happy to be overruled.

The example names a real constant — `SISEKORD_4_P_11_7_FIRST_ESCALATING_BREACH_DAY` in `TrackingDifferenceNotifier`, with the same citation opening the `V1_253` and `V1_256` migrations — so this is a question about the citation everywhere, not just the line in AGENTS.md.

I'd keep the Estonian, and I think the rule this line states argues for keeping it. `Sisekord nr 4 p 11.7` is the document's title, not a description of it. The whole point of moving rationale into the identifier is that whoever hits the constant can find the source; anglicising the title to something like `INTERNAL_RULES_4_11_7_…` breaks the one string that matches the document, the migration comments and the rule as it is actually cited, so the citation degrades into a paraphrase that greps for nothing.

English everywhere else in the identifier still holds — `FIRST_ESCALATING_BREACH_DAY` carries the meaning, and the Estonian is confined to the proper noun, the same way we would not translate an ISIN or a statute number.

If you would rather have consistency than traceability, I will rename — but then across the two migration comments as well, not only here, so the citation does not end up spelled two ways.

PR #2014, opened 2026-09-18
Measure a fee base against the NAV that charged it

taavipertman-glitch:
Four fixes to `FEE_BASE_COMPLETENESS`, found by reading three standing FAILs in production and working out which were real.

### The yardstick could move without the subject

`nav_report` keeps every calculation for a date and the check read the newest one, published or not. An ad-hoc recalculation lands there before anything decides whether it goes out — and it re-accrues the day's fees on its own components on the way — so the check was measuring an accrual against a NAV nobody was ever charged on. The two admin calculations on 2026-09-02 described in #2012 did exactly that: run without a fund parameter, they left an accrual priced before the day's prices had arrived, and then supplied the yardstick it was judged by.

The sum now selects the latest **published** calculation, ordered by `published_at` like `findPublishedNavPerUnit` and for the same reason: a backdated recalculation carries a lower id but a later publication. A date with rows but nothing published is `NOT_RUN` rather than a deviation — the report email is what marks rows published, so a failed send leaves a calculated, accrued day with no NAV to reconcile against, and `NOT_RUN` keeps that day in the window until one exists.

### The deviation amount was not a comparable quantity

Both fees are charged on components of one calculation, so one wrong calculation diverged once per fee type and the two were added: TKF100 reported `55,386.50` for a `27,693.25` error. Summed signed, they also cancelled across days — TUK75 reported `7,891,925.89` for deviations totalling `9,893,397.32`. A day is now worth its widest absolute fee-type deviation, summed over days; the per-type detail stays in the message.

It remains a **base** deviation, not a fee deviation — the fee at stake is `base × rate / 365`.

### A missing month end lowered the depot tier band instead of leaving it unknown

The anchor date is chosen publication-blind while the assets are now read published-only, so a month end whose NAV never went out would read as a fund worth nothing and drop the combined basis a band. It now fails soft the way a missing rate and a missing tier already do. A fund with no calculation at all on or before the anchor still contributes zero — it had not launched.

### One cutover date cannot describe every fund

Each fund's accrual for a position date is written by its own NAV run the next morning, and the run that moved the depot base from netovara to gross aktiva landed between those runs:

| fund | first day on the gross base |
|---|---|
| TKF100 | 2026-08-17 |
| TUV100 | 2026-08-17 |
| TUK75 | 2026-08-18 |
| TUK00 | never showed the difference — no day in range with both payables and pending redemptions |

So the date in force reported TUK75's 17 August as short by exactly that day's non-fee liabilities — a finding nothing could clear, which pinned the fund's entire daily window through `findOldestUnresolvedDailyDeviationDate`. Days before the cutover are now left uncompared rather than held to the superseded definition, and the cutover moves to `2026-08-18`. Nothing is lost: every depot accrual in that range is zero at a zero rate. `expectedBase` drops to the single definition the Depooleping names.

### For the reviewer

- **`requireNonNull` is gone** from `checkDivergence`; a fee type with no expectation is skipped. Bounded: only the cutover rule can produce that state, since every other fee type falls through `expectedBase`'s else-branch and always gets one.
- **`NOT_RUN` detail key renamed** `daysWithoutNavReport` → `daysWithoutPublishedNav`. Nothing in the repo reads it; a saved Metabase question digging into `result -> 'findings' -> 'details'` would.
- **Interacts with #2013**, which keys alert re-announcement on `deviation_amount`. No file overlap, but whichever merges second re-announces the standing failures once as the arithmetic changes. Not a new deviation.
- **One commit, not four.** The changes interleave inside `FeeBaseCompletenessChecker` — the null-skip and the max-absolute share a hunk — so they could not be split without re-editing. Happy to split if you would rather review them separately.

### Verification

`./gradlew test` green over the full suite; `./gradlew build -x test` green (Spotless, NullAway, PMD). New coverage: two repository tests (ignores a newer unpublished calculation, picks the most recently published), three `FundNavQueryService` tests, two integration tests (an unpublished recalculation does not turn a correct fee base into a FAIL; a never-published day is `NOT_RUN`), four checker tests for the deviation arithmetic and the cutover, and one `DepotFeeCalculator` test that the tier is never priced off an incomplete basis.


JordanValdma (approved):

JordanValdma:
Merged to master through #2013, which was stacked on this branch and contains all of its commits (head e368cfaaa is an ancestor of master). This PR was auto-closed by GitHub when #2003's branch was deleted, so it could not be reopened or retargeted.

